PR #5, opened 2022-01-17
Moved state connector logic into separate component

bogdanprodanj:
The goal of this PR is to move state connector logic into a separate component in order to simplify the overall code complexity

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/deckarep/golang-set v1.7.1
 	github.com/ethereum/go-ethereum v1.10.12
 	github.com/fjl/memsize v0.0.0-20190710130421-bcb5799ab5e5
-	github.com/flare-foundation/flare v0.5.1-0.20220219175705-0f29b2ba0e48
+	github.com/flare-foundation/flare v0.5.1-0.20220219193849-77148a5607bd
 	github.com/gballet/go-libpcsclite v0.0.0-20191108122812-4678299bea08
 	github.com/google/uuid v1.1.5
 	github.com/gorilla/rpc v1.2.0
--- a/go.sum
+++ b/go.sum
@@ -159,8 +159,8 @@ github.com/flare-foundation/coreth v0.3.1-0.20220219134233-0c0a76526e9d/go.mod h
 github.com/flare-foundation/flare v0.5.0/go.mod h1:ai++PWjLrLOZYheY7qutUJbReWsJSe5XO47QcIQQW0U=
 github.com/flare-foundation/flare v0.5.1-0.20220218223919-053cf219b47c/go.mod h1:s7Pdz389BY8S4R59VEO43qfO59NGVOANIjbqGm99FcE=
 github.com/flare-foundation/flare v0.5.1-0.20220219122715-d2993195fa0a/go.mod h1:9xMiQ+VgZQdBPxJoNAiwfrkFHFKreBn17yKMZiUbXx0=
-github.com/flare-foundation/flare v0.5.1-0.20220219175705-0f29b2ba0e48 h1:jQn4bzQA4j+bu+P4t1dWikU0f/4BKNVE0k9yMutdLq8=
-github.com/flare-foundation/flare v0.5.1-0.20220219175705-0f29b2ba0e48/go.mod h1:9YWA/n6TVkmRE4DrjXpumNcQJ8dp0miSbAYKw4sJ26Q=
+github.com/flare-foundation/flare v0.5.1-0.20220219193849-77148a5607bd h1:z4CxDAL1VjGSkZRgwYclddVYTUJGRTW3KnCX7R1Lbxk=
+github.com/flare-foundation/flare v0.5.1-0.20220219193849-77148a5607bd/go.mod h1:9YWA/n6TVkmRE4DrjXpumNcQJ8dp0miSbAYKw4sJ26Q=
 github.com/fogleman/gg v1.2.1-0.20190220221249-0403632d5b90/go.mod h1:R/bRT+9gY/C5z7JzPU0zXsXHKM4/ayA+zqcVNZzPa1k=
 github.com/fogleman/gg v1.3.0/go.mod h1:R/bRT+9gY/C5z7JzPU0zXsXHKM4/ayA+zqcVNZzPa1k=
 github.com/fortytw2/leaktest v1.3.0/go.mod h1:jDsjWgpAGjm2CA7WthBh/CdZYEPF31XHquHwclZch5g=
--- a/plugin/evm/version.go
+++ b/plugin/evm/version.go
@@ -11,7 +11,7 @@ var (
 	// GitCommit is set by the build script
 	GitCommit string
 	// Version is the version of Coreth
-	Version string = "v0.5.1"
+	Version string = "v0.3.1"
 )
 
 func init() {


core/state_connector.go
```diff
@@ -14,219 +15,204 @@ import (
...
+	return attestors, nil
 }
 
-func GetStateConnectorCoinbaseSignalAddr(chainID *big.Int, blockTime *big.Int) common.Address {
-	switch {
-	default:
-		return common.HexToAddress("0x000000000000000000000000000000000000dEaD")
+func (c *stateConnector) attestationResult(attestor common.Address, instructions []byte) (string, error) {
+	merkleRootHash, _, err := c.caller.Call(vm.AccountRef(attestor), c.to(), instructions, 20000, big.NewInt(0))
+	return hex.EncodeToString(merkleRootHash), err
```
Ullaakut, line 160: I'd argue that we can drop the `merkle` here without losing any meaning. It's a root hash, doesn't really matter if it's the root hash for a merkle tree or even a patricia merkle one IMO

core/state_connector.go
```diff
@@ -14,219 +15,204 @@ import (
...
 	}
+	return *c.msg.To()
 }
 
-func SubmitAttestationSelector(chainID *big.Int, blockTime *big.Int) []byte {
+func stateConnectorCoinbaseSignalAddr(chainID *big.Int, blockTime *big.Int) common.Address {
 	switch {
 	default:
-		return []byte{0xcf, 0xd1, 0xfd, 0xad}
+		return common.HexToAddress("0x000000000000000000000000000000000000dEaD")
```
Ullaakut, line 175: Maybe this should be in a const?

core/state_connector.go
```diff
@@ -14,219 +15,204 @@ import (
...
-			attestors = append(attestors, common.BytesToAddress(priceProvidersBytes[i*32:(i+1)*32]))
-		}
-		return attestors, nil
-	}
-}
-
-func GetEnvAttestationProviders(attestorType string) []common.Address {
-	envAttestationProvidersString := os.Getenv(attestorType + "_ATTESTATION_PROVIDERS")
+func localAttestors() []common.Address {
+	envAttestationProvidersString := os.Getenv("LOCAL_ATTESTATION_PROVIDERS")
```
Ullaakut, line 208: This too

core/state_connector_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"errors"
+	"fmt"
+	"math/big"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/flare-foundation/coreth/core/types"
+	"github.com/flare-foundation/coreth/core/vm"
```
Ullaakut, line 13: Why group ethereum and flare packages together?

Usually we group the following way:

* standard library
* external packages
* internal packages

And optionally if there is a strong close dependency (like for Flow packages for the Flow DPS, or here ethereum packages I assume) we have a group specifically for it between external and internal packages.

But maybe it's done this way here for consistency with other files in the repo?

core/state_connector_test.go
```diff
@@ -0,0 +1,153 @@
...
+	case m.testCase == "default_attestors_error_2":
+		switch addr {
+		case common.HexToAddress(GetPrioritisedFTSOContract(nil)):
+			return []byte("voter"), 0, nil
+		case common.BytesToAddress([]byte("voter")):
+			return nil, 0, errors.New("list price provider error")
+		}
+	}
+	return nil, 0, errors.New("undefined test case")
+}
```
Ullaakut, line 86: I really don't like this approach. IMO we should not rely on a string that describes the test to trigger a different mechanism. Instead, the [mock should have a function as its attribute, which can be overloaded by each test](https://optakt.github.io/code/testing/#mocks).

This way, the mock itself is simple, and all of the complexity that describes the test functionally and _specifies_ it so to speak, is contain within the test itself.

core/state_connector_test.go
```diff
@@ -0,0 +1,153 @@
...
+			context: vm.BlockContext{
+				Coinbase: common.BigToAddress(big.NewInt(1000)),
+			},
+		}
+		c := newConnector(mockSCC, mockMsg)
+		currentRoundNumber := []byte("222")
+		err := c.finalisePreviousRound(big.NewInt(10), big.NewInt(10), currentRoundNumber)
+		assert.NoError(t, err)
+		assert.Equal(t, common.BigToAddress(big.NewInt(1000)), mockSCC.context.Coinbase, "coinbase address should be changed to the original address")
+	})
```
Ullaakut, line 109: Why are those tests not parallelized? It seems to me like they don't interact in any way, so they should run in parallel.

core/state_connector_test.go
```diff
@@ -0,0 +1,153 @@
...
+			testCase: "happy_path",
+			msg:      mockMsg,
+			context: vm.BlockContext{
+				Coinbase: common.BigToAddress(big.NewInt(1000)),
+			},
+		}
+		c := newConnector(mockSCC, mockMsg)
+		currentRoundNumber := []byte("222")
+		err := c.finalisePreviousRound(big.NewInt(10), big.NewInt(10), currentRoundNumber)
+		assert.NoError(t, err)
```
Ullaakut, line 107: This should be a `require` instead of `assert` IMO, since we don't really care about asserting the coinbase address if there was an error here.

core/state_connector_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return nil, 0, errors.New("undefined test case")
+}
+
+func (m *mockStateConnectorCaller) WithBlockContext(bc vm.BlockContext) { m.context = bc }
+
+func (m *mockStateConnectorCaller) BlockContext() vm.BlockContext { return m.context }
+
+func TestStateTransition_FinalisePreviousRound(t *testing.T) {
+	t.Run("happy_path", func(t *testing.T) {
```
Ullaakut, line 93: We usually describe the subtest with sentences instead of snake-cased names. Here we would usually call this `nominal case`, and other cases `handles voter whitelister error` and so on. It's more descriptive than `default_attestors_error_1`

core/state_connector_test.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		mockSCC := &mockStateConnectorCaller{
+			testCase: "finalized_data_error",
+			msg:      mockMsg,
+			context: vm.BlockContext{
+				Coinbase: common.BigToAddress(big.NewInt(1000)),
+			},
+		}
+		c := newConnector(mockSCC, mockMsg)
+		currentRoundNumber := []byte("222")
```
Ullaakut, line 148: Currently you used the same current round number everywhere, so if there's no need to make it different (nothing to test with it) it can just be a var defined at the top of the function once, and reused by each subtest.

core/state_connector_test.go
```diff
@@ -0,0 +1,153 @@
...
+		mockMsg := &mockMessage{
+			from: common.BytesToAddress([]byte("finalized_data_error")),
+		}
+		mockSCC := &mockStateConnectorCaller{
+			testCase: "finalized_data_error",
+			msg:      mockMsg,
+			context: vm.BlockContext{
+				Coinbase: common.BigToAddress(big.NewInt(1000)),
+			},
+		}
```
Ullaakut, line 146: These could be created using a helper function, with just 2 arguments:

```suggestion
		msg, scc := buildStateConnectorMock("finalized_data_error", 1000)
		scc.CallFunc = func() { // Override call function for specific test case.
```

And buildStateConnectorMock would be as simple as

```go
func buildStateConnectorMock(from string, address int) (*mockMessage, *mockStateConnectorCaller) {
	msg := mockMessage{
		from: common.BytesToAddress([]byte(from)),
	}
	scc := mockStateConnectorCaller{
		testCase: "finalized_data_error",
		msg:      mockMsg,
		context: vm.BlockContext{
			Coinbase: common.BigToAddress(big.NewInt(address)),
		},
	}
	return &msg, &scc
}
```

core/state_transition.go
```diff
@@ -363,14 +375,18 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
+		if vmerr == nil &&
+			*msg.To() == stateConnectorContract(chainID, timestamp) &&
+			len(st.data) >= 36 &&
+			len(ret) == 32 &&
+			isStateConnectorActivated(chainID, timestamp) &&
+			// 4 first bytes corresponds to the function name
+			bytes.Equal(st.data[0:4], submitAttestationSelector(chainID, timestamp)) &&
+			// determine whether this is read-only call
+			binary.BigEndian.Uint64(ret[24:32]) > 0 {
+			err = st.stateConnector.finalisePreviousRound(chainID, timestamp, st.data[4:36])
```
Ullaakut, line 387: Lines 379 to 386 and line 387 should not be at  the same level of indentation IMO, it's confusing 🤔 Was this code go-fmted?

core/state_transition.go
```diff
@@ -363,14 +375,18 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
+			if err != nil {
+				log.Warn("error finalising state connector round", "error", err)
```
Ullaakut, line 389: Isn't this effectively the same as

```go
log.Warn("error finalising state connector round error", err)
```

? If so, I'd rather reformat it like such

```suggestion
				log.Warn("error finalising state connector round:", err)
```

core/state_transition.go
```diff
@@ -432,6 +448,39 @@ func (st *StateTransition) refundGas(apricotPhase1 bool) {
 	st.gp.AddGas(st.gas)
 }
 
+func stateConnectorContract(chainID *big.Int, blockTime *big.Int) common.Address {
+	switch {
+	case isStateConnectorActivated(chainID, blockTime) && chainID.Cmp(songbirdChainID) == 0:
+		return songbirdStateConnectorAddress
+	default:
+		return common.HexToAddress("0x1000000000000000000000000000000000000001")
```
Ullaakut, line 456: Should be in a const

core/state_transition.go
```diff
@@ -432,6 +448,39 @@ func (st *StateTransition) refundGas(apricotPhase1 bool) {
...
+
+func isTestingChain(chainID *big.Int) bool {
+	return chainID.Cmp(flareChainID) != 0 && chainID.Cmp(songbirdChainID) != 0
+}
+
+func submitAttestationSelector(chainID *big.Int, blockTime *big.Int) []byte {
+	switch {
+	default:
+		return []byte{0xcf, 0xd1, 0xfd, 0xad}
+	}
```
Ullaakut, line 481: No point in having a switch here. It can be added later if needed but here it just makes the code less clear for no good  reason.

core/state_transition.go
bogdanprodan-okta: there should be an even number of arguments provided, that's why it's `"error", err`

bogdanprodanj: There should be an even number of arguments provided, that's why it's `"error", err`

Ullaakut: Ah okay my bad, I thought this was the standard `log` package but it's actually a structured logger from ethereum. Makes sense then!

core/state_transition.go
Ullaakut: Actually, since it is indeed formatted, I have an alternate suggestion @bogdanprodanj: Put this logic in an appropriately named function.

```go
func shouldFinalize(...) bool {
	if *msg.To() != stateConnectorContract(chainID, timestamp) {
		return false
	}
	if len(st.data) < 36 {
		return false
	}
	if len(ret) != 32 {
		return false
	}
	if !isStateConnectorActivated(chainID, timestamp) {
		return false
	}
	if !bytes.Equal(st.data[0:4], submitAttestationSelector(chainID, timestamp)) {
		return flase
	}
	if binary.BigEndian.Uint64(ret[24:32]) <= 0 {
		return false
	}

	return true
}
```

The only issue with that is that it needs access to the following variables: `chainID`, `timestamp`, `msg`, `st` and `ret`.

Ullaakut (approved):

core/state_connector.go
```diff
@@ -14,219 +15,196 @@ import (
...
+	var finalityReached bool
+	if len(localAttestors) > 0 {
+		localAttestationVotes := c.countAttestations(localAttestors, instructions)
+		if defaultAttestationVotes.reachedMajority && localAttestationVotes.reachedMajority && defaultAttestationVotes.majorityDecision == localAttestationVotes.majorityDecision {
+			finalityReached = true
+		} else if defaultAttestationVotes.reachedMajority && defaultAttestationVotes.majorityDecision != localAttestationVotes.majorityDecision {
+			// FIXME Make a back-up of the current state database, because this node is about to branch from the default set
+		}
+	} else if defaultAttestationVotes.reachedMajority {
+		finalityReached = true
```
ricardomgoncalves, line 67: You could put this in a function to make the code more clear, and just check for the finalityReached, WDYT?

Ullaakut:
@bogdanprodanj What's the status of this PR?

bogdanprodanj:
> @bogdanprodanj What's the status of this PR?

If @awfm9 doesn't have any objections, it's ready to be merged

Ullaakut:
You should maybe add @awfm9 as reviewer of the PR so that he has it in his notifications/Pull Request tab, if he uses those 😄

awfm9:
I'll take a look soon.

core/state_connector.go
```diff
@@ -174,6 +178,8 @@ func (c *stateConnector) to() common.Address {
 	return *c.msg.To()
 }
 
+// envAttestors returns list of attestors from environment variable using provided key.
+// Returns empty list of value does not exist.
```
Ullaakut, line 182:
```suggestion
// Returns an empty list of value if the key does not exist in the environment.
```

core/state_connector.go
```diff
@@ -187,10 +193,61 @@ func envAttestors(key string) []common.Address {
 	return attestors
 }
 
-func attestationSelector() []byte {
-	return []byte{0x29, 0xbe, 0x4d, 0xb2}
+// unused 'blockTime' might be used for hard forks in the future.
+func stateConnectorContract(chainID *big.Int, blockTime *big.Int) common.Address {
```
Ullaakut, line 197: If it's unused I'd personally rather not have it now, and add it the day we need it. WDYT?

core/state_connector.go
```diff
@@ -187,10 +193,61 @@ func envAttestors(key string) []common.Address {
...
+	}
+}
+
+// unused 'chainID' and 'blockTime' might be used for hard forks in the future.
+func submitAttestationSelectorFunc(chainID *big.Int, blockTime *big.Int) []byte {
+	switch {
+	default:
+		return []byte{0xcf, 0xd1, 0xfd, 0xad}
+	}
+}
```
Ullaakut, line 229: Same here, for now I'd rather have it just return `[]byte{0xcf, 0xd1, 0xfd, 0xad}` without the switch or unused parameters, but it's up to you.

core/state_connector.go
```diff
@@ -187,10 +193,61 @@ func envAttestors(key string) []common.Address {
...
+
+// unused 'chainID' and 'blockTime' might be used for hard forks in the future.
+func stateConnectorCoinbaseSignalAddr(chainID *big.Int, blockTime *big.Int) common.Address {
+	switch {
+	default:
+		return common.HexToAddress("0x000000000000000000000000000000000000dEaD")
```
Ullaakut, line 235: It's quite uncommon to mix lowercase and uppercase characters in an hexadecimal string. I assume the source you copied it from had it formatted like this? I'd rather make it fully uppercase for readability if that's ok with you

awfm9: This is the checksum functionality of Ethereum addresses. The checksum is stored in the upper/lower case difference.

Ullaakut (approved):

PR #8, opened 2022-01-28
Extra gRPC endpoint to get validators

emailtovamos:
This code is very rough right now which just works based on a dummy smart contract which is included in the Flare repo's genesis_local file.
This will be replaced by the actual smart contract when it gets ready. 
After that this evm call in this current code change will be changed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/deckarep/golang-set v1.7.1
 	github.com/ethereum/go-ethereum v1.10.12
 	github.com/fjl/memsize v0.0.0-20190710130421-bcb5799ab5e5
-	github.com/flare-foundation/flare v0.4.7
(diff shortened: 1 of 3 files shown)


Ullaakut:
Apparently I had a month old review in progress on this PR which I never finished 😄 Why was this closed/never finished?

emailtovamos:
Hey @Ullaakut , sorry I closed it without writing the reason. 

This was a placeholder to setup the gRPC endpoint assuming that the mechanism to get the validators and the corresponding weights will be there at smart contract level. But there has been a lot of changes with regards to the approach from then and some calculations now need to be done at the Coreth level.  

I made the changes and the code was not that clean and also it was getting an import cycle error after taking the changes from the updated master. So I decided to start from scratch and make it a bit cleaner. This is there currently in https://github.com/flare-foundation/coreth/pull/10 

So that is reviewable and has most of the core logic. But it is yet to be tested so I expect further changes.

PR #13, opened 2022-03-17
Add hard fork activation and state connector activation

awfm9:


core/state_connector.go
```diff
@@ -21,12 +22,8 @@ const (
 )
 
 var (
-	costonChainID   = new(big.Int).SetUint64(16) // https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-16.json
-	songbirdChainID = new(big.Int).SetUint64(19) // https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-19.json
-	flareChainID    = new(big.Int).SetUint64(14) // https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-14.json
-
 	costonActivationTime   = big.NewInt(time.Date(2022, time.February, 25, 17, 0, 0, 0, time.UTC).Unix())
-	songbirdActivationTime = big.NewInt(time.Date(2200, time.January, 1, 0, 0, 0, 0, time.UTC).Unix())
+	songbirdActivationTime = big.NewInt(time.Date(2022, time.March, 28, 15, 0, 0, 0, time.UTC).Unix())
```
Ullaakut: Why/how is this related to the rest of the PR? I thought songbird was already activated so I feel like I'm missing something here.

emailtovamos, line 26: In Flare repo this is: `constants.SongbirdID: time.Date(2022, time.March, 28, 14, 0, 0, 0, time.UTC),`

I would have expected these to be the same or Coreth first before Flare or am I missing something?

plugin/evm/ftso_system.go
```diff
@@ -114,28 +114,37 @@ func (f *FTSOSystem) Contracts(hash common.Hash) (FTSOContracts, error) {
...
-	}
-
 	var managerAddress common.Address
-	err = snap.OnContract(f.submitter).Execute(ManagerAddress).Decode(&managerAddress)
+	err := snap.OnContract(f.submitter).Execute(ManagerAddress).Decode(&managerAddress)
 	if err != nil {
 		return FTSOContracts{}, fmt.Errorf("could not get manager address: %w", err)
 	}
 
+	empty := common.Address{}
```
bogdanprodanj, line 123:
```suggestion
        // check if managerAddress is non-empty
	if managerAddress == common.Address{} {
		return FTSOContracts{}, errFTSONotDeployed
	}
```

core/vm/contracts.go
```diff
@@ -130,6 +130,8 @@ func init() {
 // ActivePrecompiles returns the precompiles enabled with the current configuration.
 func ActivePrecompiles(rules params.Rules) []common.Address {
 	switch {
+	case rules.IsFlareHardFork1:
```
emailtovamos, line 133: Two cases returning the same value. So they can be merged?

emailtovamos, line 133: Two cases returning the same value. So they can be merged?

plugin/evm/validators_default.go
```diff
@@ -0,0 +1,74 @@
...
+	case chainID.Cmp(params.SongbirdChainID) == 0:
+		nodeIDs = songbirdNodeIDs
+	case chainID.Cmp(params.FlareChainID) == 0:
+		nodeIDs = flareNodeIDs
+	default:
+		nodeIDs = strings.Split(os.Getenv("DEFAULT_VALIDATORS"), ",")
+	}
+
+	if len(nodeIDs) == 0 {
+		return nil, fmt.Errorf("no default validators set")
```
bogdanprodanj, line 61: we can use var here
`errEmptyValidatorsSet = errors.New("no default validators set")`

plugin/evm/ftso_system.go
awfm9: You can't write that, the syntax won't work because of the curly braces.

plugin/evm/validators_default.go
awfm9: That's not needed, we are only using sentinel errors for two specific cases because we don't want to treat those as an error.

core/state_connector.go
awfm9: This is the state connector activation time, not the hard fork.

core/vm/contracts.go
awfm9: No, they are evaluated in order.

PR #14, opened 2022-03-18
Added integration tests for evm calls

bogdanprodanj:


plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,91 @@
...
+func TestEVMCall_Execute(t *testing.T) {
+
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+
+	auth, err := bind.NewKeyedTransactorWithChainID(key, big.NewInt(1337))
+	require.NoError(t, err)
+
+	balance := new(big.Int)
+	balance.SetString("100000000000000000000", 10)
```
awfm9, line 29: You can also use `SetUint64` and use a number literal or actual `uint64` variable.

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,91 @@
...
...10088578063f56256c71461010b575b600080fd5b6100726004803603602081101561005c57600080fd5b8101908080359060200190929190505050610143565b6040518082815260200191505060405180910390f35b61009061015b565b6040518080602001828103825283818151815260200191508051906020019080838360005b838110156100d05780820151818401526020810190506100b5565b50505050905090810190601f1680156100fd5780820380516001836020036101000a031916815260200191505b509250505060405180910390f35b6101416004803603604081101561012157600080fd5b8101908080359060200190929190803590602001909291905050506101f9565b005b60016020528060005260406000206000915090505481565b60008054600181600116156101000203166002900480601f0160208091040260200160405190810160405280929190818152602001828054600181600116156101000203166002900480156101f15780601f106101c6576101008083540402835291602001916101f1565b820191906000526020600020905b8154815290600101906020018083116101d457829003601f168201915b505050505081565b8060016000848152602001908152602001600020819055507fe79e73da417710ae99aa2088575580a60415d359acfad9cdd3382d59c80281d48282604051808381526020018281526020019250505060405180910390a1505056fea2646970667358221220a7cedd091a0b3c6a40ba59ec51143153fbaa5bf19f0d8d0fd927e7f6dc84974164736f6c63430007060033`
```
awfm9, line 39: It's not an `abiBin`; it would be more correct to talk about the `contractABI` and the `contractBin`. The nomenclature in the Ethereum community is `contractByteCode`.

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,91 @@
...
+	auth, err := bind.NewKeyedTransactorWithChainID(key, big.NewInt(1337))
+	require.NoError(t, err)
+
+	balance := new(big.Int)
+	balance.SetString("100000000000000000000", 10)
+	alloc := make(core.GenesisAlloc)
+	alloc[auth.From] = core.GenesisAccount{
+		Balance: balance,
+	}
+	gasLimit := uint64(9999999)
```
awfm9, line 34: You probably want to just use `math.MaxUint64` here, as all of our calls to the EVM are read-only with the same unlimited gas.

awfm9 (commented):
Already looks quite solid! I would want to extend them significantly to cover many more cases, but this is exactly the most critical part to test.

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,91 @@
+package evm
+
```
Ullaakut, line 2: If these are integration tests and that they turn out to be slower/heavier than unit tests, we should make sure to add a build tag for it. This way when we have a CI for it, we could choose to not run the integration tests in some cases to save time.

```suggestion
package evm

//go:build integration
```

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,91 @@
...
+	contract := EVMContract{
+		address: contractAddr,
+		abi:     parsed,
+	}
+
+	blockChain := be.Blockchain()
+
+	evmBind := BindEVM(blockChain)
+
+	t.Run("success_call", func(t *testing.T) {
```
Ullaakut, line 58:
```suggestion
	t.Run("nominal case", func(t *testing.T) {
```

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,91 @@
...
+		be.Commit(true)
+
+		header = blockChain.GetHeaderByNumber(2)
+
+		ret = evmBind.AtBlock(header.Hash()).OnContract(contract).Execute("items", [32]byte{1})
+		require.NoError(t, ret.err)
+		require.Len(t, ret.values, 1)
+	})
+
+	t.Run("invalid_method", func(t *testing.T) {
```
Ullaakut, line 73:
```suggestion
	t.Run("handles invalid method", func(t *testing.T) {
```

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,91 @@
...
+		header := blockChain.GetHeaderByNumber(2)
+		ret := evmBind.AtBlock(header.Hash()).OnContract(contract).Execute("setItem2", [32]byte{1}, [32]byte{1})
+		assert.EqualError(t, ret.err, "could not pack parameters: method 'setItem2' not found")
+	})
+
+	t.Run("invalid_params", func(t *testing.T) {
```
Ullaakut, line 79:
```suggestion
	t.Run("handles invalid parameters", func(t *testing.T) {
```

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,91 @@
...
+		header := blockChain.GetHeaderByNumber(2)
+		ret := evmBind.AtBlock(header.Hash()).OnContract(contract).Execute("setItem", [32]byte{1}, [32]byte{1}, [32]byte{1})
+		assert.EqualError(t, ret.err, "could not pack parameters: argument count mismatch: got 3 for 2")
+	})
+
+	t.Run("invalid_header_number", func(t *testing.T) {
```
Ullaakut, line 85:
```suggestion
	t.Run("handles invalid header number", func(t *testing.T) {
```

Ullaakut (approved):
Small nitpicks, besides that LGTM 👍

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,95 @@
...
+	auth, err := bind.NewKeyedTransactorWithChainID(key, big.NewInt(1337))
+	require.NoError(t, err)
+
+	balance := new(big.Int)
+	balance.SetUint64(math.MaxUint64)
+	alloc := make(core.GenesisAlloc)
+	alloc[auth.From] = core.GenesisAccount{
+		Balance: balance,
+	}
+	var gasLimit uint64 = math.MaxUint64
```
awfm9, line 38: `gasLimit := uint64(math.MaxUint64)` is more idiomatic here.

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,95 @@
...
+func TestEVMCall_Execute(t *testing.T) {
+
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+
+	auth, err := bind.NewKeyedTransactorWithChainID(key, big.NewInt(1337))
+	require.NoError(t, err)
+
+	balance := new(big.Int)
+	balance.SetUint64(math.MaxUint64)
```
awfm9, line 33: `balance := big.NewInt(0).SetUint64(math.MaxUint64)` is more idiomatic here.

Ullaakut (approved):

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,95 @@
...
...10088578063f56256c71461010b575b600080fd5b6100726004803603602081101561005c57600080fd5b8101908080359060200190929190505050610143565b6040518082815260200191505060405180910390f35b61009061015b565b6040518080602001828103825283818151815260200191508051906020019080838360005b838110156100d05780820151818401526020810190506100b5565b50505050905090810190601f1680156100fd5780820380516001836020036101000a031916815260200191505b509250505060405180910390f35b6101416004803603604081101561012157600080fd5b8101908080359060200190929190803590602001909291905050506101f9565b005b60016020528060005260406000206000915090505481565b60008054600181600116156101000203166002900480601f0160208091040260200160405190810160405280929190818152602001828054600181600116156101000203166002900480156101f15780601f106101c6576101008083540402835291602001916101f1565b820191906000526020600020905b8154815290600101906020018083116101d457829003601f168201915b505050505081565b8060016000848152602001908152602001600020819055507fe79e73da417710ae99aa2088575580a60415d359acfad9cdd3382d59c80281d48282604051808381526020018281526020019250505060405180910390a1505056fea2646970667358221220a7cedd091a0b3c6a40ba59ec51143153fbaa5bf19f0d8d0fd927e7f6dc84974164736f6c63430007060033`
```
awfm9, line 43: You might want to put these as constants somewhere in their own file, where we can add more of them as needed.

plugin/evm/evm_call_test.go
```diff
@@ -0,0 +1,95 @@
...
+		ret := evmBind.AtBlock(header.Hash()).OnContract(contract).Execute("setItem", [32]byte{1}, [32]byte{1})
+		require.NoError(t, ret.err)
+
+		be.Commit(true)
+
+		header = blockChain.GetHeaderByNumber(2)
+
+		ret = evmBind.AtBlock(header.Hash()).OnContract(contract).Execute("items", [32]byte{1})
+		require.NoError(t, ret.err)
+		require.Len(t, ret.values, 1)
```
awfm9, line 74: I'm missing something here that checks that the value was actually correctly set on the smart contract? We just check there is a return value, but not its actual value.

plugin/evm/ftso_system_test.go
```diff
@@ -0,0 +1,221 @@
...
+	testAbiVotePowerJSON = `[{"inputs":[{"internalType":"uint256","name":"_vpInt","type":"uint256"}],"stateMutability":"nonpayable","type":"constructor"},{"inputs":[{"internalType":"address","name":"provider","type":"address"}],"name":"votePowerOf","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"vpInt","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"}]`
+	testAbiVotePowerBin  = `0x608060405234801561001057600080fd5b506040516101493803806101498339818101604052602081101561003357600080fd5b8101908080519060200190929190505050806000819055505060ef8061005a6000396000f3fe6080604052348015600f57600080fd5b506004361060325760003560e01c8063142d10181460375780638966b43214608c575b600080fd5b607660048036036020811015604b57600080fd5b81019080803573ffffffffffffffffffffffffffffffffffffffff16906020019092919050505060a8565b6040518082815260200191505060405180910390f35b609260b3565b6040518082815260200191505060405180910390f35b600080549050919050565b6000548156fea264697066735822122043d0eedff37b596f65fc5d3f9b98f8444bac96fcf123c979b81ece56c2991d2664736f6c63430007060033`
```
awfm9, line 23: The JSON and byte code could also all go into a separate file, as constants.

plugin/evm/ftso_system_test.go
```diff
@@ -0,0 +1,221 @@
...
+	testAbiVotePower, _  = abi.JSON(strings.NewReader(testAbiVotePowerJSON))
```
awfm9, line 24: ABIs should be decoded as part of an `init` function or a test setup function, so that we can actually check they happened correctly. Otherwise, we might have silent errors.

plugin/evm/ftso_system_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+		// VotePower contract
+		vpInt := big.NewInt(100)
+		votePowerAddr := deployTestContract(t, auth, be, testAbiVotePower, testAbiVotePowerBin, vpInt)
+
+		// Wnat contract
+		wnatAddr := deployTestContract(t, auth, be, testAbiWnat, testAbiWnatBin, votePowerAddr)
+
+		// Reward contract
+		unclaimedReward := big.NewInt(100)
```
awfm9, line 74: This should be a different number. Otherwise, we have no way of checking whether we returned unclaimed rewards or vote power.

plugin/evm/ftso_system_test.go
```diff
@@ -0,0 +1,221 @@
...
+		rewardAddr := deployTestContract(t, auth, be, testAbiReward, testAbiRewardBin, wnatAddr, unclaimedReward)
+
+		// Manager contract
+		rewardEpochDurationSeconds := big.NewInt(100)
+		rewardEpochPowerHeight := big.NewInt(100)
+		rewardEpochStartHeight := big.NewInt(100)
+		rewardEpochStartTime := big.NewInt(100)
```
awfm9, line 81: Same here. These all need to be unique values, so we are sure we are distinguishing them correctly when retrieving.

plugin/evm/ftso_system_test.go
```diff
@@ -0,0 +1,221 @@
...
+		supportedIndices := []*big.Int{index1, index2}
+		ftsoRegistryAddr := deployTestContract(t, auth, be, testAbiFtsoRegistry, testAbiFtsoRegistryBin, supportedIndices)
+
+		// Whitelist contract
+		indices := []*big.Int{big.NewInt(123)}
+		priceProvidersAddresses := []common.Address{common.BigToAddress(big.NewInt(222))}
+		voterWhitelisterAddr := deployTestContract(t, auth, be, testAbiVoterWhitelister, testAbiVoterWhitelisterBin, indices, priceProvidersAddresses)
+
+		// Validation contract
+		providerAddresses := []common.Address{common.BigToAddress(big.NewInt(222))}
```
awfm9, line 97: Again, all values should be unique. It's a good idea to actually define all values at the very top of the test, and then use named variables in the initialization logic.

awfm9 (approved):
Looks good :+1:

I added some minor comments that you could address before merging.

plugin/evm/evm_call_test.go
bogdanprodanj: Since we are only reading from smart contracts and not writing, I guess I'll need some extra code to actually write something to the smart contract. Correct me if I'm wrong.

P.S. I've tested it using standard eth libs to generate the smart contract, send some eth there and it worked fine. But here it just doesn't work :(

plugin/evm/cotracts_test.go
```diff
@@ -0,0 +1,133 @@
+//go:build integration
```
Ullaakut, line 1: Careful, the name of this file has a typo in it. Also I guess it should have a copyright header? Not sure if they're 100% consistent with their headers in coreth.

plugin/evm/ftso_system_test.go
```diff
@@ -1,101 +1,65 @@
...
-		// VotePower contract
 		vpInt := big.NewInt(100)
+
+		unclaimedReward := big.NewInt(200)
+
+		rewardEpochDurationSeconds := big.NewInt(301)
+		rewardEpochPowerHeight := big.NewInt(302)
+		rewardEpochStartHeight := big.NewInt(303)
+		rewardEpochStartTime := big.NewInt(304)
+		rewardEpochsStartTs := big.NewInt(1) // this should be less than header.Time()
```
Ullaakut, line 31: What is `header.Time()` in this context? There's no header variable I can see

bogdanprodanj: it's just a note for this line of the code `plugin/evm/ftso_system.go:144`

plugin/evm/cotracts_test.go
bogdanprodanj: Yeah, I'm not sure about the headers either

PR #19, opened 2022-03-20
default validators test

emailtovamos:


plugin/evm/validators_default_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/flare-foundation/coreth/params"
+)
+
+func TestGetDefaultValidators(t *testing.T) {
+	tests := []struct {
+		chainID            *big.Int
+		weightOfAValidator uint64
+		numberOfValidators int
```
Ullaakut, line 17:
```suggestion
		name    string
		chainID *big.Int
		weight  uint64
		amount  int
```

plugin/evm/validators_default_test.go
```diff
@@ -0,0 +1,41 @@
...
+	for _, test := range tests {
+		validatorsMap, err := getDefaultValidators(test.chainID)
+		require.NoError(t, err)
+		assert.Equal(t, test.numberOfValidators, len(validatorsMap))
+		totalWeight := uint64(0)
+		for _, u := range validatorsMap {
+			totalWeight += u
+		}
+		assert.Equal(t, uint64(len(validatorsMap))*test.weightOfAValidator, totalWeight)
+	}
```
Ullaakut, line 40:
```suggestion
	for _, test := range tests {
		test := test

		t.Run(test.name, func(t *testing.T) {
			t.Parallel()

			got, err := getDefaultValidators(test.chainID)
			require.NoError(t, err)

			assert.Len(t, got, test.amount)
			gotWeight := uint64(0)
			for _, v := range got {
				gotWeight += v
			}
			assert.Equal(t, uint64(test.amount)*test.weight, totalWeight)
		}
	}
```

plugin/evm/validators_default_test.go
```diff
@@ -0,0 +1,41 @@
...
+		{
+			chainID:            params.CostonChainID,
+			weightOfAValidator: costonValidatorWeight,
+			numberOfValidators: 5,
+		},
+		{
+			chainID:            params.SongbirdChainID,
+			weightOfAValidator: songbirdValidatorWeight,
+			numberOfValidators: 20,
+		},
```
Ullaakut, line 28: We should also have a test with the FlareChainID and one with a custom validator set. For the test with the custom validator set, since it involves an environment variable, it could make sense to do it outside of the table-driven-test here, and instead write this test in a separate test function.

Ullaakut (changes requested):

plugin/evm/validators_default_test.go
emailtovamos: Ok, but for FlareChainID an error is expected as the list is nil right now. Should I write a separate one for that as well or have checks inside the loop itself to do `require.Error()`  ?

Ullaakut: Yep, you can add an attribute called `wantErr` of type `require.ErrorAssertionFunc` and set it to `require.NoError` for other tests, and `require.Error` for this one :)

plugin/evm/validators_default_test.go
```diff
@@ -0,0 +1,69 @@
...
+	require.NoError(t, err)
+	got, err := getDefaultValidators(big.NewInt(444))
+	require.NoError(t, err)
+	assert.Len(t, got, 2)
+	gotWeight := uint64(0)
+	for _, v := range got {
+		gotWeight += v
+	}
+	assert.Equal(t, uint64(2)*customValidatorWeight, gotWeight)
+}
```
Ullaakut, line 69:
```suggestion
func TestGetDefaultValidators_CustomValidators(t *testing.T) {
	err := os.Setenv("CUSTOM_VALIDATORS", "NodeID-3M9KVT6ixi4gVMisbm5TnPXYXgFN5LHuv,NodeID-NnX4fajAmyvpL9RLfheNdc47FKKDuQW8i")
	require.NoError(t, err)

	got, err := getDefaultValidators(big.NewInt(444))
	require.NoError(t, err)

	assert.Len(t, got, 2)
	var gotWeight uint64
	for _, v := range got {
		gotWeight += v
	}
	assert.Equal(t, uint64(2)*customValidatorWeight, gotWeight)
}
```

plugin/evm/validators_default_test.go
```diff
@@ -0,0 +1,69 @@
...
+			got, err := getDefaultValidators(test.chainID)
+			test.wantErr(t, err)
+			assert.Len(t, got, test.amount)
+			gotWeight := uint64(0)
+			for _, v := range got {
+				gotWeight += v
+			}
+			assert.Equal(t, uint64(test.amount)*test.weight, gotWeight)
+		})
+	}
```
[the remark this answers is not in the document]
Ullaakut: Indeed, my bad 😄

Ullaakut (approved):
LGTM, just needs a rebase

emailtovamos:
This was merged but yet I can't see it in the branch. Not sure why.

Ullaakut:
I forgot about this. I also can't see it, I'm looking into it now.

PR #21, opened 2022-03-20
evm: add unit test to validators

AnNiran:


plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+		for i := 0; i < 25; i++ {
+			slots := rand.Intn(1000)
+			cacheOption := WithCacheSize(uint(slots))
+			cacheOption(&testCfg)
+
+			assert.Equal(t, testCfg.CacheSize, uint(slots))
+		}
+	})
+}
```
Ullaakut, line 40: I'm not sure why your test checks the returned function's type 🤔 Why would it change upon calling the function?

Testing this option func should be as simple as the following IMO:

```suggestion
func TestWithCacheSize(t *testing.T) {
	wantCacheSize := 42

	cfg := CacheConfig{}
	WithCacheSize(wantCacheSize)(&cfg)

	assert.Equal(t, wantCacheSize, cfg.CacheSize)
}
```

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,40 @@
+package evm
```
Ullaakut, line 1: Shouldn't this file also test `ValidatorsCache.ByEpoch`?

plugin/evm/validators_ftso_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"math/rand"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWithRootDEgree(t *testing.T) {
```
Ullaakut, line 11:
```suggestion
func TestWithRootDegree(t *testing.T) {
```

plugin/evm/validators_ftso_test.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		for i := 0; i < 25; i++ {
+			degree := rand.Intn(1000)
+			ftsoOption := WithRootDegree(uint(degree))
+			ftsoOption(&testCfg)
+
+			assert.Equal(t, testCfg.RootDegree, uint(degree))
+		}
+	})
```
Ullaakut, line 39: Same suggestion here. Test should be simplified.

plugin/evm/validators_manager_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"encoding/binary"
+	"math/rand"
+	"testing"
+
+	"github.com/flare-foundation/flare/ids"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
```
Ullaakut, line 11:
```suggestion
import (
	"encoding/binary"
	"math/rand"
	"testing"

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"

	"github.com/flare-foundation/flare/ids"
)
```

plugin/evm/validators_manager_test.go
```diff
@@ -0,0 +1,65 @@
...
+		shortIDKeys = append(shortIDKeys, randArray())
+	}
+
+	testValidatorsMap := map[ids.ShortID]uint64{
+		ids.ShortID(shortIDKeys[0]): testUint64Slice[0],
+		ids.ShortID(shortIDKeys[1]): testUint64Slice[1],
+		ids.ShortID(shortIDKeys[2]): testUint64Slice[2],
+	}
+
+	set, err := toSet(testValidatorsMap)
```
Ullaakut, line 45: Should be simplified this way IMO

```suggestion
	var (
		ids []ids.ShortID
		weights []uint64
	)
	for i := 0; i < 3; i++ {
		weight, err := randUint64()
		require.NoError(t, err)
		weights = append(weights, weight)

		id := randID()
		ids = append(ids, id)
	}

	validators := map[ids.ShortID]uint64{
		ids[0]: weights[0],
		ids[1]: weights[1],
		ids[2]: weights[2],
	}

	set, err := toSet(validators)
```

plugin/evm/validators_manager_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func randArray() [20]byte {
+	data := [20]byte{}
+	rand.Read(data[:])
+
+	return data
+}
```
Ullaakut, line 18:
```suggestion
func randShortID() ids.ShortID {
	id := ids.ShortID{}
	rand.Read(id[:])

	return id
}
```

plugin/evm/validators_manager_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+func randUint64() (uint64, error) {
+	data := make([]byte, 8)
+	_, err := rand.Read(data)
+	return binary.LittleEndian.Uint64(data), err
+}
```
Ullaakut, line 24: Why not just use [rand.Uint64](https://pkg.go.dev/math/rand#Uint64)? 🤔

plugin/evm/validators_manager_test.go
```diff
@@ -0,0 +1,65 @@
...
+		ids.ShortID(shortIDKeys[1]): testUint64Slice[1],
+		ids.ShortID(shortIDKeys[2]): testUint64Slice[2],
+	}
+
+	set, err := toSet(testValidatorsMap)
+	require.NoError(t, err)
+
+	assert.Equal(t, set.Contains(shortIDKeys[0]), true)
+	assert.Equal(t, set.Contains(shortIDKeys[1]), true)
+	assert.Equal(t, set.Contains(shortIDKeys[2]), true)
```
Ullaakut, line 50:
```suggestion
	assert.True(t, set.Contains(shortIDKeys[0]))
	assert.True(t, set.Contains(shortIDKeys[1]))
	assert.True(t, set.Contains(shortIDKeys[2]))
```

plugin/evm/validators_manager_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	weight, ok = set.GetWeight(shortIDKeys[1])
+	assert.Equal(t, weight, testUint64Slice[1])
+	assert.Equal(t, ok, true)
+
+	weight, ok = set.GetWeight(shortIDKeys[2])
+	assert.Equal(t, weight, testUint64Slice[2])
+	assert.Equal(t, ok, true)
+
+	assert.Equal(t, set.Len(), len(testValidatorsMap))
```
Ullaakut, line 64:
```suggestion
	assert.Len(t, set, len(validators))
```

Ullaakut (changes requested):

plugin/evm/validators_cache_test.go
AnNiran: Yes, I am currently working on this, since those function require much more research to do from me.
This PR is created with the work done until now

plugin/evm/validators_cache_test.go
AnNiran: Why you removed the tested behavior ?

plugin/evm/validators_manager_test.go
AnNiran: Because I forgot, thanks!

plugin/evm/validators_ftso_test.go
AnNiran: Fixed https://github.com/flare-foundation/coreth/pull/21/commits/b5d59c2475c2f081d8ccb493da8daedbd87067ff#diff-5d4bfd4a98a9e3484f1bb6efb6fee69a6b752832d96c0ed354bec7be522ce905R11
Thanks!

plugin/evm/validators_manager_test.go
AnNiran: This is simplified and moved because this function was moved to vm 
https://github.com/flare-foundation/coreth/pull/21/commits/b5d59c2475c2f081d8ccb493da8daedbd87067ff#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32R484
Thanks!

plugin/evm/validators_manager_test.go
AnNiran: OKay I fixed this https://github.com/flare-foundation/coreth/pull/21/commits/b5d59c2475c2f081d8ccb493da8daedbd87067ff#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R12
but now I see I made it randID(), should also work I think, yes?

plugin/evm/validators_manager_test.go
AnNiran: Here -> https://github.com/flare-foundation/coreth/pull/21/commits/b5d59c2475c2f081d8ccb493da8daedbd87067ff#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32R490
its using simply rand.Uint64 now
Thanks!

plugin/evm/validators_manager_test.go
AnNiran: This is all fixed here https://github.com/flare-foundation/coreth/pull/21/commits/b5d59c2475c2f081d8ccb493da8daedbd87067ff#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32R503-R524
and also added a new test case because the function was changed

plugin/evm/validators_manager_test.go
AnNiran: Corrected https://github.com/flare-foundation/coreth/pull/21/commits/b5d59c2475c2f081d8ccb493da8daedbd87067ff#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32R523
Thanks!

plugin/evm/validators_cache_test.go
AnNiran: Yes, corrected https://github.com/flare-foundation/coreth/pull/21/commits/4b455472b4f4659e87bf3e6827c1bd26078e71ee#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7L19
Thanks!

plugin/evm/validators_manager_test.go
AnNiran: This import list has been removed, because the function from this file has been moved to vm_test as per my comment https://github.com/flare-foundation/coreth/pull/21#discussion_r830626604

plugin/evm/validators_ftso_test.go
AnNiran: Please note that I removed the initial version here also:
https://github.com/flare-foundation/coreth/pull/21/commits/5c646cdde6cb04d81b0d03b20e2bcd58846aeb6c#diff-5d4bfd4a98a9e3484f1bb6efb6fee69a6b752832d96c0ed354bec7be522ce905
Thanks!

plugin/evm/validators_manager_test.go
Ullaakut: It's fine yes :)

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,87 @@
...
+	validatorsCache := NewValidatorsCache(cacheTestData, WithCacheSize(32))
+	for k, v := range cacheTestData {
+		validatorsCache.cache.Add(k, v)
+	}
+
+	t.Run("NonExistingKey", func(t *testing.T) {
+		res, err := validatorsCache.ByEpoch(uint64(7))
+		require.NoError(t, err)
+		assert.Len(t, res, 0)
+	})
```
Ullaakut, line 69:
```suggestion
	t.Run("handles missing key", func(t *testing.T) {
		res, err := validatorsCache.ByEpoch(uint64(7))
		require.NoError(t, err)
		assert.Len(t, res, 0)
	})
```

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,87 @@
...
+			result, err := validatorsCache.ByEpoch(epoch)
+			require.NoError(t, err)
+
+			value, ok := result[idsList[pos]]
+			assert.True(t, ok)
+
+			expected := cacheTestData[epoch][idsList[pos]]
+			assert.Equal(t, value, expected)
+		}
+	})
```
Ullaakut, line 86: This test should be defined first since it is the nominal case. Also be careful, `assert` functions take the expected value first and the assessed second

```suggestion
	t.Run("nominal case", func(t *testing.T) {
		for i, epoch := range epochs {
			result, err := validatorsCache.ByEpoch(epoch)
			require.NoError(t, err)

			pos := numValidators - 1
			if i == 1 {
				pos = (numValidators / 2) - 1
			}
			value, ok := result[idsList[pos]]
			assert.True(t, ok)

			want:= cacheTestData[epoch][idsList[pos]]
			assert.Equal(t, want, value)
		}
	})
```

plugin/evm/validators_manager_test.go
```diff
@@ -0,0 +1 @@
+package evm
```
Ullaakut, line 1: ?

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"math"
+	"math/rand"
+	"reflect"
+	"testing"
+
+	"github.com/flare-foundation/flare/ids"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
```
Ullaakut, line 12:
```suggestion
import (
	"math"
	"math/rand"
	"reflect"
	"testing"

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"

	"github.com/flare-foundation/flare/ids"
)
```

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,88 @@
...
+			continue
+		}
+
+		normTestData[epochs[1]] = map[ids.ShortID]uint64{
+			idsList[i]: weights[i],
+		}
+		copyNormTestData[epochs[1]] = map[ids.ShortID]uint64{
+			idsList[i]: weights[i],
+		}
+	}
```
Ullaakut, line 45: Same suggestion here.

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	t.Run("ExistingKey", func(t *testing.T) {
+		for _, epoch := range epochs {
+			result, err := normalizer.ByEpoch(epoch)
+			require.NoError(t, err)
+
+			var weightedNTData = map[ids.ShortID]uint64{}
+			for id, _ := range result {
+				weightedNTData[id] = copyNormTestData[epoch][id]
+			}
```
Ullaakut, line 67: I'm not fond of this solution tbh. If the point is to check that the weight ratio is the same, we can just compute it in advance and define it as a variable at the top of the test, and then use it only for the assertion. WDYT?

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+			weighted := calcWeightRatio(weightedNTData)
+			assert.Equal(t, len(weighted), len(result))
```
Ullaakut, line 70:
```suggestion
			assert.Len(t, result, len(weighted))
```

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,88 @@
...
+			assert.True(t, reflect.DeepEqual(weighted, result))
```
Ullaakut, line 71: You can use [`assert.ElementsMatch`](https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch) for this I believe

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,88 @@
...
+		totalWeight += weight
+	}
+
+	ratio := math.MaxInt64 / totalWeight
+	for validator, weight := range validators {
+		validators[validator] = weight * ratio
+	}
+
+	return validators
+}
```
Ullaakut, line 88: So here if I get this right, we basically copy the original logic and incorporate it into the test? 😄 I think this is a mistake as it means that if someone modifies the original code they might not realize they need to update it here as well, and also it's not a very DRY approach.

Again I think it would be easier to just hardcode the expected `map[ids.ShortID]uint64` at the top of the file 👍

plugin/evm/vm_test.go
```diff
@@ -477,3 +480,53 @@ func TestConfigureLogLevel(t *testing.T) {
...
+		idsList = append(idsList, id)
+	}
+
+	validators := map[ids.ShortID]uint64{
+		idsList[0]: weights[0],
+		idsList[1]: weights[1],
+		idsList[2]: weights[2],
+	}
+
+	t.Run("Validators", func(t *testing.T) {
```
Ullaakut, line 503:
```suggestion
	t.Run("nominal case", func(t *testing.T) {
```

plugin/evm/vm_test.go
```diff
@@ -477,3 +480,53 @@ func TestConfigureLogLevel(t *testing.T) {
...
+		assert.Equal(t, ok, true)
+
+		weight, ok = set.GetWeight(idsList[2])
+		assert.Equal(t, weight, weights[2])
+		assert.Equal(t, ok, true)
+
+		assert.Len(t, set, len(validators))
+	})
+
+	t.Run("Error", func(t *testing.T) {
```
Ullaakut, line 526:
```suggestion
	t.Run("handles given error", func(t *testing.T) {
```

plugin/evm/vm_test.go
```diff
@@ -477,3 +480,53 @@ func TestConfigureLogLevel(t *testing.T) {
 		})
 	}
 }
+
+func TestToSet(t *testing.T) {
```
Ullaakut, line 484: In this test we're also missing a subtest for the case where `set.AddWeight` fails

Ullaakut (changes requested):

plugin/evm/validators_normalizer_test.go
AnNiran: We are not checking if the weight ratio is the same.
We are taking the old ratio to make parallel calculations that match the ones the `ByEpoch` method is doing, in order to
arrive at the same result of _some_ number of map values

plugin/evm/validators_normalizer_test.go
AnNiran: This is not allowed by assert.Len() here

plugin/evm/validators_normalizer_test.go
AnNiran: Why to use it since it is using reflect.DeepEqual itself also?

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,87 @@
+package evm
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/flare-foundation/flare/ids"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
```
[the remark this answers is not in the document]
AnNiran: This has been reorganized https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R10

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,87 @@
...
+func (testData) ByEpoch(uint64) (map[ids.ShortID]uint64, error) {
+	return map[ids.ShortID]uint64{}, nil
+}
+
+func randID() ids.ShortID {
+	id := ids.ShortID{}
+	rand.Read(id[:])
+
+	return id
+}
```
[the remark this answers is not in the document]
AnNiran: Moved here
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7L18
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R79

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,87 @@
...
+			cacheTestData[epochs[0]] = map[ids.ShortID]uint64{
+				idsList[i]: weights[i],
+			}
+			continue
+		}
+
+		cacheTestData[epochs[1]] = map[ids.ShortID]uint64{
+			idsList[i]: weights[i],
+		}
+	}
```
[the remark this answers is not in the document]
AnNiran: Okay,
Changed it in that way
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R38

plugin/evm/validators_cache_test.go
AnNiran: Fixed:
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R72

plugin/evm/validators_cache_test.go
AnNiran: Moved it to first place and fixed:
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R57

AnNiran: Moved it to first place and fixed:
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R57

plugin/evm/validators_manager_test.go
AnNiran: Yes, this is the testing file that was left from some previous refactoring and that most probably will be used again.
Deleted for now

plugin/evm/validators_normalizer_test.go
AnNiran: This is also reorganized
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-3bb410a6d1f259ada3678375ee08a83bd2a8bf2e93945e5c92b92a6f9d99e796R11

plugin/evm/validators_normalizer_test.go
Ullaakut: Because it was made for this 😄 Instead of combining two methods (check if it's true that they are equal) we just use a single one that does the same thing clearly (check if they are equal)

plugin/evm/vm_test.go
AnNiran: This is changed also:
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32R503

plugin/evm/vm_test.go
AnNiran: Also fixed:
https://github.com/flare-foundation/coreth/pull/21/commits/795a19e455a7a1b0b6a427200d807bfb6ec6d3dc#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32R526

plugin/evm/validators_normalizer_test.go
Ullaakut: If we do this, then we should also have a test for that extra function, but sure that sounds fine to me. As to not using hardcoded values, I agree when it's simple things but when most of the test consists in defining the variables we use I don't like it 😬 The way we did it in Flow DPS is that we had a [`mocks` package in which we had functions to generate test values randomly](https://github.com/optakt/flow-dps/blob/master/testing/mocks/generic.go) (but also deterministically, so that tests can be reproduced).

WDYT of doing something similar here? Where we have a helper function that generates both the validator set with IDs and weights, and computes the weight ratio for the generated values at the same time?

plugin/evm/validators_normalizer_test.go
Ullaakut: Why not? `assert.Len` takes any interface that is a slice or a map (can be given to the `len` builtin) and an int, so `len(weighted)` should be fine 🤔

plugin/evm/validators_normalizer_test.go
AnNiran: "If we do this, then we should also have a test for that extra function, but sure that sounds fine to me. " - yes, I was thinking the same thing.
Yes, I was thinking about separating the generation into a function last night. It can generate values accordingly for the tests :+1: 
Yes, that will be used for all the tests in validators_ , and not only, tests

AnNiran: :handshake:  :smile:

plugin/evm/validators_normalizer_test.go
AnNiran: Okay :smile:

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	validatorsCache := NewValidatorsCache(cacheTestData, WithCacheSize(32))
+	for k, v := range cacheTestData {
+		validatorsCache.cache.Add(k, v)
+	}
+
+	t.Run("nominal case", func(t *testing.T) {
+		for i, epoch := range epochs {
```
emailtovamos, line 58: Where exactly is this `epochs` defined? I can't find it.

Ullaakut: In one of my suggestions I proposed that we remove the slice that was `epochs := []uint64{1, 2}` and I guess this is a remainder of it. I'll make a suggestion to improve this loop.

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,84 @@
...
+			assert.True(t, ok)
+			want:= cacheTestData[epoch][idsList[pos]]
+			assert.Equal(t, want, value)
+		}
+	})
+
+	t.Run("handles missing key", func(t *testing.T) {
+		res, err := validatorsCache.ByEpoch(uint64(7))
+		require.NoError(t, err)
+		assert.Len(t, res, 0)
```
Ullaakut, line 75: This is not needed IMO. In Go when a function returns an error, its other return values should be assumed to be invalid data that could lead to undefined behavior (similar to how in C++ you're not supposed to be dereferencing null or dangling pointers).

In practice, in most cases, we just return a `nil` pointer or a zero value, but still because of that we don't need to check it in an error test IMO.

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	validatorsCache := NewValidatorsCache(normTestData, WithCacheSize(128))
+	for k, v := range normTestData {
+		validatorsCache.cache.Add(k, v)
+	}
+	normalizer := NewValidatorsNormalizer(validatorsCache)
+
+	t.Run("nominal case", func(t *testing.T) {
+		for _, epoch := range epochs {
```
Ullaakut, line 56: Same here, `epochs` no longer exists. See previous thread

emailtovamos:
Does this branch need a rebasing or something?

AnNiran:
> Does this branch need a rebasing or something?

No, I already rebased it locally, what do you mean?

emailtovamos:
Ok. I saw the number of files changed is `53` which is why I asked.

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/flare-foundation/flare/ids"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type testData map[uint64]map[ids.ShortID]uint64
+
+func (testData) ByEpoch(uint64) (map[ids.ShortID]uint64, error) {
+	return map[ids.ShortID]uint64{}, nil
+}
```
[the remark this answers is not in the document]
AnNiran: Because I only need to satisfy the interface.
And the data type I need to create the test is not the returned ```map[ids.ShortId]uint64``` but ```map[uint64]map[ids.ShortId]uint64```

plugin/evm/vm_test.go
AnNiran: Not sure when this function returns error though https://github.com/flare-foundation/flare/blob/master/snow/validators/set.go#L205
:thinking:

PR #22, opened 2022-03-20
Epochs manager unit test

emailtovamos:


plugin/evm/epochs_manager_test.go
```diff
@@ -0,0 +1,75 @@
...
+				EndTime:   big.NewInt(time.Date(2022, time.January, 3, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
+			}
+			epochDetails3 := EpochDetails{
+				StartTime: big.NewInt(time.Date(2022, time.January, 3, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
+				EndTime:   big.NewInt(time.Date(2022, time.January, 4, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
+			}
+			epochDetails4 := EpochDetails{
+				StartTime: big.NewInt(time.Date(2022, time.January, 4, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
+				EndTime:   big.NewInt(time.Date(2022, time.December, 4, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
+			}
```
Ullaakut, line 41: These should not be defined in this func but instead be at the top of the testing function. This way the body of the function can focus on the actual logic.

plugin/evm/epochs_manager_test.go
```diff
@@ -0,0 +1,75 @@
...
+			switch epoch {
+			case 0:
+				return epochDetails1, nil
+			case 1:
+				return epochDetails2, nil
+			case 2:
+				return epochDetails3, nil
+			default:
+				return epochDetails4, nil
```
Ullaakut, line 53: Instead of doing this and making a mock with logic to handle multiple tests, please just overwrite the function with one that returns exactly what the test needs in each subtest.

plugin/evm/epochs_manager_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	timeJan1 := big.NewInt(time.Date(2022, time.January, 1, 17, 0, 0, 0, time.UTC).Unix()).Uint64()
+	epoch, err = epochManager.ByTimestamp(timeJan1)
+	require.NoError(t, err)
+	assert.Equal(t, uint64(0), epoch)
+
+	timeJan2 := big.NewInt(time.Date(2022, time.January, 2, 17, 0, 0, 0, time.UTC).Unix()).Uint64()
+	epoch, err = epochManager.ByTimestamp(timeJan2)
+	require.NoError(t, err)
+	assert.Equal(t, uint64(1), epoch)
```
Ullaakut, line 74: Please have subtests that test the following:

* The epoch mock returns an epoch that matches the requested timestamp directly
* The epoch mock returns a bunch of epochs that do not match what we need, and after a while it returns the one we want (when doing this, we need to check that the mock was called the expected amount of times)
* The epoch mock returns an error

Ullaakut (changes requested):

plugin/evm/epochs_manager_test.go
emailtovamos: `DetailsFunc()` has to be called multiple times so that the `epoch` settles at the right range. I am not sure how returning one single thing will achieve this? Maybe if you can give an example it will be clearer.

Ullaakut: There is only one subtest where we need to call it multiple times, and only in this specific subtest we'll want to have something like

```go
testDetails := []EpochDetails{
			{
				StartTime: big.NewInt(time.Date(2022, time.January, 1, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
				EndTime:   big.NewInt(time.Date(2022, time.January, 2, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
			},
			{
				StartTime: big.NewInt(time.Date(2022, time.January, 2, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
				EndTime:   big.NewInt(time.Date(2022, time.January, 3, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
			},
			{
				StartTime: big.NewInt(time.Date(2022, time.January, 3, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
				EndTime:   big.NewInt(time.Date(2022, time.January, 4, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
			},
			{
				StartTime: big.NewInt(time.Date(2022, time.January, 4, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
				EndTime:   big.NewInt(time.Date(2022, time.December, 4, 0, 0, 0, 0, time.UTC).Unix()).Uint64(),
			}
}

t.Run("finds correct epoch details among multiple calls", func(t *testing.T) {
	var called int
	mock := &testEpochs{
		DetailsFunc: func(epoch uint64) (EpochDetails, error) {
			called++
			return testDetails[i], nil
		},
	}

	// ...
}
```

Do you see what I mean? :)

emailtovamos: I understand better now but still not fully clear as I don't know what `i` is in your above suggestion. 
I am almost done for today, so if you have time later, feel free to push your suggestions. Thanks :)

Ullaakut: My bad, I edited the snippet, I meant for it to be `epoch`, and to use an epoch that matches the index in the slice.

Ullaakut:
Will push a commit to address my suggestions, please do not work on this PR in the meantime to avoid conflicts :)

emailtovamos:
Haha thanks for letting me know :)

plugin/evm/epochs_manager_test.go
```diff
@@ -19,57 +19,114 @@ func (t *testEpochs) Details(epoch uint64) (EpochDetails, error) {
...
+		assert.Equal(t, 8, callsCount)
+
+		// Request central epoch. Expect 2 more iterations forwards.
+		epoch, err = epochManager.ByTimestamp(epochs[2].StartTime)
+		require.NoError(t, err)
+		assert.Equal(t, uint64(1), epoch)
+		assert.Equal(t, 10, callsCount)
+	})
+
+	t.Run("request first epoch", func(t *testing.T) {
```
emailtovamos, line 56: I am wondering in which kind of scenario, the "nominal case" will work but this test will fail.

Ullaakut: There is indeed some overlap between the tests, but it doesn't matter much IMO. We want to make sure that requesting each of the boundaries works (first and last epoch) as well as intermediate values and going back and forth (nominal case).

Ullaakut (approved):

PR #24, opened 2022-03-22
Test validators transitioner

emailtovamos:
This is a work in progress. 3 basic checks are included so far but it needs some more in depth checks to be fully complete. 

TODO:

- [x] Do more with providers checks especially the actual transition phase.
- [x] Check for weights
- [x] Non nil error check
- [x] Basic cache usage check
- [x] Non nil default validators

### DIFF
--- a/plugin/evm/validators_transitioner_test.go
+++ b/plugin/evm/validators_transitioner_test.go
@@ -1,0 +1,161 @@
+package evm
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/flare-foundation/flare/ids"
+)
+
+type testValidatorRetriever struct {
+	ByEpochFunc func(epoch uint64) (map[ids.ShortID]uint64, error)
+}
+
+func (t *testValidatorRetriever) ByEpoch(epoch uint64) (map[ids.ShortID]uint64, error) {
+	return t.ByEpochFunc(epoch)
+}
+
+func TestValidatorsTransitioner_ByEpoch(t *testing.T) {
+
+	t.Run("nominal case 1: includes check for recursion and cache usage", func(t *testing.T) {
+		t.Parallel()
+
+		calls := 0
+		testEpoch := uint64(10)
+		providers := fakeProviders(testEpoch)
+
+		mock := &testValidatorRetriever{
+			ByEpochFunc: func(epoch uint64) (map[ids.ShortID]uint64, error) {
+				calls++
+				return providers, nil
+			},
+		}
+
+		validatorsTransitioner := NewValidatorsTransitioner(nil, mock)
+
+		got, err := validatorsTransitioner.ByEpoch(testEpoch)
+		require.NoError(t, err)
+		assert.Equal(t, providers, got)
+		assert.Equal(t, 10, calls)
+
+		// calls is expected to increase by 1 only because the cache would be used
+		got, err = validatorsTransitioner.ByEpoch(testEpoch)
+		require.NoError(t, err)
+		assert.Equal(t, providers, got)
+		assert.Equal(t, 11, calls)
+
+		mock.ByEpochFunc = func(epoch uint64) (map[ids.ShortID]uint64, error) {
+			calls++
+			return providers, nil
+		}
+
+		got, err = validatorsTransitioner.ByEpoch(testEpoch + 1)
+		require.NoError(t, err)
+		assert.Equal(t, providers, got)
+		assert.Equal(t, 13, calls)
+	})
+
+	t.Run("nominal case 2: non-nil default validators to reach the end of the function", func(t *testing.T) {
+		t.Parallel()
+
+		testEpoch := uint64(10)
+		providers := fakeProviders(testEpoch)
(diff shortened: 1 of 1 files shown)


plugin/evm/validators_transitioner_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+		got, err := validatorsTransitioner.ByEpoch(testEpoch)
+		require.NoError(t, err)
+		assert.Equal(t, providers, got)
+
+		var totalWeight uint64
+		for _, u := range got {
+			totalWeight += u
+		}
+		assert.Equal(t, uint64(45), totalWeight)
```
Ullaakut, line 85:
```suggestion
		testEpoch := uint64(10)
		wantWeight := uint64(45)
		providers := fakeProviders(testEpoch)
		// here we give non-nil validators so that the later part of the ByEpoch() function can be reached by control and therefore tested
		validators := fakeValidators(testEpoch)

		mock := &testValidatorRetriever{
			ByEpochFunc: func(epoch uint64) (map[ids.ShortID]uint64, error) {
				return providers, nil
			},
		}

		validatorsTransitioner := NewValidatorsTransitioner(validators, mock)

		got, err := validatorsTransitioner.ByEpoch(testEpoch)
		require.NoError(t, err)
		assert.Equal(t, providers, got)

		var totalWeight uint64
		for _, u := range got {
			totalWeight += u
		}
		assert.Equal(t, wantWeight , totalWeight)
```

Small nitpick but it's usually better to define "magic" expected output variables at the top of the test so that the reader has a clear idea of the ins and outs of the test while they read it IMO (but it's debatable, that's just my personal preference 😄)

plugin/evm/validators_transitioner_test.go
```diff
@@ -0,0 +1,161 @@
...
+		got, err := validatorsTransitioner.ByEpoch(testEpoch)
+		require.NoError(t, err)
+		assert.Equal(t, validators, got)
+	})
+
+	t.Run("case for non nil error", func(t *testing.T) {
+		t.Parallel()
+
+		mock := &testValidatorRetriever{
+			ByEpochFunc: func(epoch uint64) (map[ids.ShortID]uint64, error) {
```
Ullaakut, line 112:
```suggestion
			ByEpochFunc: func(uint64) (map[ids.ShortID]uint64, error) {
```

Here since we don't use `epoch` we can explicitly ignore it, but that's a nitpick as well no worries :)

plugin/evm/validators_transitioner_test.go
```diff
@@ -0,0 +1,161 @@
...
+				return nil, fmt.Errorf("error in getting valirators by epoch: %d", epoch)
+			},
+		}
+		validatorsTransitioner := NewValidatorsTransitioner(nil, mock)
+		providers, err := validatorsTransitioner.ByEpoch(1)
+		require.Error(t, err)
+		assert.Nil(t, providers)
```
Ullaakut, line 119:
```suggestion
		}

		validatorsTransitioner := NewValidatorsTransitioner(nil, mock)

		_, err := validatorsTransitioner.ByEpoch(1)
		require.Error(t, err)
```

Some more nitpicks:

* I tend to have newlines between the following test sections
	* Test variable definitions (inputs, expected outputs)
	* Mock definitions (newline between each mock)
	* Test subject definition (in this case the transitioner)
	* Call on test subject and assertions
* If we check that we get an error, we should not make any assertions on the other returned parameter. When an error is returned, the other parameter should be expected to be an invalid arbitrary value

plugin/evm/validators_transitioner_test.go
```diff
@@ -0,0 +1,161 @@
...
+	})
+}
+
+func fakeProviders(epoch uint64) map[ids.ShortID]uint64 {
+
+	providers := make(map[ids.ShortID]uint64)
+
+	for i := 0; i < int(epoch); i++ {
+		providers[ids.ShortID{byte(i)}] = uint64(i)
+	}
```
Ullaakut, line 146:
```suggestion
	providers := make(map[ids.ShortID]uint64)
	for i := 0; i < int(epoch); i++ {
		providers[ids.ShortID{byte(i)}] = uint64(i)
	}
```

When we define a variable and "fill" it in a loop, the definition should cuddle with the loop.  See [WSL Rules](https://github.com/bombsimon/wsl/blob/master/doc/rules.md) for some interesting thoughts on that. Some of them we (and specifically @awfm9) disagree with and do not follow, though, so to be read while taking it with a grain of salt 😄

plugin/evm/validators_transitioner_test.go
emailtovamos: I agree with this for the other tests. But in this test in particular, I am including the `epoch` inside the error message so it is not really unused :)

Ullaakut: But it shouldn't be used :p The error message should be just `dummy error` or something like that, since the message is irrelevant. All that matters is that `err` is not nil 👍

Ullaakut (approved):
LGTM, just one last nitpick which is a quick 2 line change :) Good job!

Ullaakut:
I just realized, I think that the https://github.com/flare-foundation/coreth/tree/validators-ftso-transition branch got merged into master before this was merged into it, so master does not have those tests 🤦 

I will fix this right now. My bad.

PR #25, opened 2022-03-22
Add unit tests to validators

AnNiran:


AnNiran:
@Ullaakut ,
Please note that I have removed my last PR due to:
- too many irrelevant (changed) commits and comments
- github did not show me the new rebase for some reason
- the discussed changes are indeed a different approach.

This PR addresses your last comments from https://github.com/flare-foundation/coreth/pull/21
and contains a more generic approach to tests - I created a separate package with the idea it to be used by other instances also (including testing of validators by timestamp) and of course it can be drastically enlarged.

I used hardcoded values at some of the places also.
Please let me know what do you think about this approach, I personally think it is better :) 
Thanks

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,66 @@
...
+	expected := map[ids.ShortID]uint64{
+		mocks.TestIdsList[2]: 515007957320493220,
+		mocks.TestIdsList[1]: 6465383072520688749,
+		mocks.TestIdsList[0]: 6185718675989118760,
+		mocks.TestIdsList[3]: 9599604527249266409,
+		mocks.TestIdsList[4]: 3417212044008417831,
+	}
+	weighted := normalizer.calcWeightRatio(validators)
+	assert.InDeltaMapValues(t, weighted, expected, 0, 0)
+}
```
Ullaakut, line 48: I guess we should also check that we do not modify the original `validators` map in here 😄

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	validatorsCache := NewValidatorsCache(normTestData, WithCacheSize(128))
+	for k, v := range normTestData {
+		validatorsCache.cache.Add(k, v)
+	}
+
+	var testCtx = snow.Context{
+		Log: &logging.Log{},
+	}
+	return NewValidatorsNormalizer(testCtx.Log, validatorsCache), copyNormTestData
```
Ullaakut, line 65:
```suggestion
	return NewValidatorsNormalizer(&logging.Log{}, validatorsCache), copyNormTestData
```

Not sure what the point is of creating a context here if we only use one of its attributes 🤔

Ullaakut (changes requested):
There are still some things I disagree with unfortunately :/ For me the main issues are:

* I'm pretty sure a major bug was introduced when the function was split to compute weight ratios separately, where now the original validator set has incorrect weights set to the ratios
* I don't think we need to have any logic in the mocks. They should IMO just be simple mocks, with simpler names, that just call the function they are given, and each test should just define the relevant function for the test, which in most cases just returns one thing without any extra logic.

Does that make sense? If anything is unclear, don't hesitate to LMK, we can do a quick pair session for this and I can provide examples of what I have in mind 👍

plugin/evm/validators_normalizer_test.go
AnNiran: Yes, corrected it here https://github.com/flare-foundation/coreth/pull/25/commits/1dd72d6d53018c84ed45547712fe9644c91bf263#diff-3bb410a6d1f259ada3678375ee08a83bd2a8bf2e93945e5c92b92a6f9d99e796R61
Thanks, missed this

plugin/evm/validators_normalizer.go
```diff
@@ -44,20 +44,24 @@ func (v *ValidatorsNormalizer) ByEpoch(epoch uint64) (map[ids.ShortID]uint64, er
...
-	for validator, weight := range validators {
-		normalized[validator] = weight * ratio
+	for val, wght := range validators {
+		validators[val] = wght * ratio
 	}
 
 	v.log.Debug("new normalized validator set: %#v", normalized)
 
-	return normalized, nil
+	return validators
```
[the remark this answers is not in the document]
AnNiran: This has been corrected here https://github.com/flare-foundation/coreth/pull/25/commits/1dd72d6d53018c84ed45547712fe9644c91bf263#diff-d83dabfc703ecd56ab6a9652ed20200b05740254974784895843edbd1e7adae4R61
Thanks!

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,43 @@
...
+	wantCacheSize := uint(42)
+
+	cfg := CacheConfig{}
+	WithCacheSize(wantCacheSize)(&cfg)
+
+	assert.Equal(t, wantCacheSize, cfg.CacheSize)
+}
+
+func TestValidatorsCache_ByEpoch(t *testing.T) {
+	cacheTestData := mocks.NewValidatorsTestDataByEpochs(6, 2)
```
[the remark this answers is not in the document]
AnNiran: I have reworked the tests using the discussed testing approach.
Please let me know what do you think, thanks!

plugin/evm/validators_cache_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/flare-foundation/coreth/tests/mocks"
+	"github.com/flare-foundation/flare/ids"
```
Ullaakut, line 10:
```suggestion
	"github.com/flare-foundation/coreth/tests/mocks"

	"github.com/flare-foundation/flare/ids"
```

I've seen in other PRs that people now split coreth and flare imports in two groups so I guess we should follow

Ullaakut:
@AnNiran Unfortunately it seems that the branch needs to be rebased again due to the recent changes on `validators-ftso-transition` :(

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	assert.Equal(t, wantCacheSize, cfg.CacheSize)
+}
+
+type TestValidatorCache struct {
+	ByEpochFunc func(uint64) (map[ids.ShortID]uint64, error)
+}
+
+func (tv TestValidatorCache) ByEpoch(e uint64) (map[ids.ShortID]uint64, error) {
+	return tv.ByEpochFunc(e)
```
Ullaakut, line 27: Nitpick.

```suggestion
func (t TestValidatorCache) ByEpoch(e uint64) (map[ids.ShortID]uint64, error) {
	return t.ByEpochFunc(e)
```

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,59 @@
...
+			assert.Equal(t, expected, result)
+		}
+	})
+
+	t.Run("missing key case", func(t *testing.T) {
+		result, err := valCache.ByEpoch(uint64(7))
+		require.NoError(t, err)
+		assert.Len(t, result, 0)
+	})
+}
```
Ullaakut, line 59: That's closer but still not exactly what I had in mind 😬 What I meant was more something along the lines of this:

* Only the nominal case contains the mocked function that returns actual legit data. The error case instead has a mock which returns errors.
	* As we discussed, in each test case, the mock logic is specified for that test specifically only.
* We add a test case for the situation where the mock returns an error.
* The calls to `mocks.RetrieverMock` are now made using named arguments that make it clear what 5 and 2 are in the context of the test
* The mock asserts that it gets the expected values as received arguments
* Instead of getting the expected result from the mock, we define it at the top of the test to make it clear to the reader what to expect. It is a bit more cumbersome than having something generic, but I'm not fond of retrieving the expected value from the mock personally.
* `tvc` is renamed to `mock` to emphasize on its purpose
* `data` is renamed to `testValidators`
* Some more small nitpicks

```suggestion
func TestValidatorsCache_ByEpoch(t *testing.T) {

	t.Run("nominal case", func(t *testing.T) {
		numValidators := 5
		numEpochs := 2
		testValidators := mocks.RetrieverMock(numValidators, numEpochs)
		want := map[ids.ShortID]uint64{
			// Fill this with the data we expect to get.
		}

		var calls int
		mock := TestValidatorCache{
			ByEpochFunc: func(e uint64) (map[ids.ShortID]uint64, error) {
				assert.Equal(t, calls, e)

				calls++
				return data[e], nil
			},
		}

		valCache := NewValidatorsCache(mock, WithCacheSize(32))
		for k, v := range testValidators {
			valCache.cache.Add(k, v)
		}

		result, err := valCache.ByEpoch(1)
		require.NoError(t, err)
		assert.Equal(t, expected, result)
	})

	t.Run("handles missing key", func(t *testing.T) {
		// I'm not sure how this test works in your case actually? Shouldn't it panic since we would attempt to get `validators[7]` in the map which does not exist?
		// I don't have time to look deeply at the code we're testing unfortunately so we might need to chat about it when I am back
	})

	t.Run("handles failure to retrieve validator by epoch", func(t *testing.T) {
		mock := TestValidatorCache{
			ByEpochFunc: func(uint64) (map[ids.ShortID]uint64, error) {
				return nil, errors.New("dummy error")
			},
		}

		valCache := NewValidatorsCache(mock, WithCacheSize(32))
	
		_, err := valCache.ByEpoch(uint64(7))
		assert.Error(t, err)
	})
}
```

plugin/evm/validators_cache_test.go
AnNiran: Here I meant `tv` standing for `test validator`, and now it seems to me that is could become `tvc` standing for `test validator cache`, so we can have also `tvn` - for test validator normalizer, etc.
Do you mean with this comment that this is inappropriate and should be simply `t`?

Ullaakut: Yes, method receivers should be a single letter in 99% of cases

plugin/evm/validators_cache_test.go
AnNiran: I have updated the tests https://github.com/flare-foundation/coreth/pull/25/commits/42101d73f668fc367336f11d03ee6a9b1b1f07f9
Please let me know what do you think

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.6.0
+	github.com/btcsuite/btcd v0.21.0-beta // indirect
```
Ullaakut, line 7: It is a bit strange that the `go.mod` file changed but not the `go.sum`. Could you run `go mod tidy` on your branch and push again?

plugin/evm/epochs_cache.go
```diff
@@ -0,0 +1,52 @@
+// (c) 2019-2021, Ava Labs, Inc. All rights reserved.
```
Ullaakut, line 1: These files do not seem related to adding unit tests to validators. Did some other changes get merged into this branch?

AnNiran: It seems to me that this came from the rebase, do you mean it should not be here?

Ullaakut: Yes, the rebase might have failed or might need to be done again? Because this PR should only include your commits.

Ullaakut:
To clarify: Closed in favor of #30

PR #29, opened 2022-03-23
Fix broken fee history

### LINKED ISSUE #28: Bad interaction between `eth_feeHistory` and FTSO provider submissions
With the introduction of AP3/EIP-1559 a couple of new RPC calls have been introduced to estimate the required gas price. One of them is `eth_feeHistory`, that is the call used by MetaMask to calculate the low/market/aggressive priority gas prices on EIP-1559 enabled chains.

FTSO providers are encouraged to use a higher gas price to prioritize their transactions, with the assumption that most of the gas is automatically refunded by the system.

When combined, the result is that the `eth_feeHistory` call can return very high priority fee values at the end/start of a price epoch, as if all recent transactions had to spend a lot of gas to be included in a block, while actually most of gas was refunded. This deceives MetaMask that proposes absurdly high (and unneeded) priority fees to send transactions, ruining all the advantages of the dynamic fees introduced in AP3.

A solution might be to ignore or to normalize the gas used by transactions sent to the PriceSubmitter, in order to have meaningful values returned by `eth_feeHistory`.

For reference, this is the MetaMask code: https://github.com/MetaMask/controllers/tree/main/src/gas

Example:

image_2022-03-23_07-49-26

awfm9:
This should fix #28.

### DIFF
--- a/eth/gasprice/feehistory.go
+++ b/eth/gasprice/feehistory.go
@@ -40,6 +40,7 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 	_ "github.com/flare-foundation/coreth/consensus/misc"
 	"github.com/flare-foundation/coreth/core/types"
+	"github.com/flare-foundation/coreth/params"
 	"github.com/flare-foundation/coreth/rpc"
 )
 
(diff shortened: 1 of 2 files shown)


awfm9:
Merged already in command line :grimacing:

Ullaakut (commented):
Posthumously approved 👍

PR #30, opened 2022-03-25
evm: add unit tests to validators.

AnNiran:


tests/mocks/validators.go
```diff
@@ -0,0 +1,81 @@
+package mocks
```
Ullaakut, line 1: Usually the package for testing helpers/mocks should be named `testing` instead for consistency with other repositories

tests/mocks/validators.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		if i == epochs-1 {
+			for ri := 0; ri < remainder; ri++ {
+				data[uint64(i)][RandID()] = rand.Uint64()
+			}
+		}
+	}
+
+	return data
+}
```
Ullaakut, line 73: Should be renamed to `GenericValidators`, `FakeValidators`, `GenerateValidators` or something like that, since here this does not return a retrieverMock at all anymore.

tests/mocks/validators.go
AnNiran: I haven't seen this until now, especially since this package is intended for generation of testing data, not for testing itself, but ok

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,175 @@
...
+			ByEpochFunc: func(e uint64) (map[ids.ShortID]uint64, error) {
+				return testValidators[e], nil
+			},
+			CalcWeightRatioFunc: func(vals map[ids.ShortID]uint64) map[ids.ShortID]uint64 {
+				for i, _ := range vals {
+					vals[i] = uint64(0)
+				}
+				calls++
+				return vals
+			},
```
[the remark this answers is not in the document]
AnNiran: Yes, this is the idea - to put zero values in the received map.
We are already giving it the copy.

tests/mocks/validators.go
AnNiran: Is your intention the package to be named as the go-testing package, also because the code is continuation of the language anyway?
In that way it could be like this:
coreth/tests/
 - validators_mocks.go
 - something_else_mocks.go
 etc.
OR
 - validators_utils.go

plugin/evm/validators_normalizer_test.go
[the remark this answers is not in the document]
Ullaakut: I don't understand the point of this mock then 🤔 It's used at the same time to return a map with zero values and to change the values of the given map? Why not just make it return `make(map[ids.ShortID]uint64)`? Why do we need to modify the received map as well?

tests/mocks/validators.go
Ullaakut: See https://github.com/optakt/flow-dps/tree/master/testing

AnNiran: WDYM? 
I see testing/mocks/mocks...?
What package name did you mean to change - the `tests` or `mocks`?

Ullaakut: There is just `testing/mocks` and `testing/helpers` in Flow  DPS. It should be the same here IMO if we wanted to have a common folder for mocks.

In this specific case, since the mocks are only used in the `plugin/evm` package though, I'm not convinced it's even needed to have them in a separate package now that I think about it, and anyway it's also not what your PR is doing since what you called mocks here is actually generic values used for testing. The mocks are defined in the `evm` package currently.

So what I would suggest is the following:

* If we need `RandID` and the function that generates test validator maps outside of this package, we create a package in `./testing/mocks` from the root of the repository with a file called `validators.go` which contains functions for generating validator maps and one named `short_ids.go` which contains the function currently named `RandID()`.
* Otherwise, we just hardcode the maps of validators that we need in the functions that need them. For example in my tests to have a fake validators map I just did this: https://github.com/flare-foundation/coreth/pull/33/files#diff-f74d73d6d9de90c767d318d06215fae9a4c4b74e3048fcc86043b6bb103131c3R49-R51 which I think is much simpler and clearer.

From what I remember, and correct me if I am wrong, the only place where we actually need a somewhat complex map of validators with multiple values is for the validators normalizer since we need to make sure the resulting map has the correct weight ratios right?

In which case, I think it would be best to just use my approach for all tests, and hardcode very simple maps everywhere, except in the relevant tests where we'd use a local test helper which generates a map with random values and expects a matching weight ratio map. WDYT?

plugin/evm/validators_normalizer_test.go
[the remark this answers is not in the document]
AnNiran: We do not need an empty map for expected value, we need a map where for the given IDs we have normalized weights zero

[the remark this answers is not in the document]
Ullaakut: OK so since we know what ids we will be given, we can hardcode that map and we do not need to modify the given one

tests/mocks/validators.go
AnNiran: Okay, `testing/mocks` is what I see in Flow-dps of course, just asked because I wasn't sure if you meant the `tests` package or `mocks` one in the first comment and yes - it makes sense `tests` to be renamed to `testing`.

My intention for using the validators helper functions outside the `evm` was to have a single place for using tests generated values across the code, instead of multiple files and packages and sub-packages.
I also see that `ids.ShortID` is used in other places outside `evm` package - `peer` and `message`, maybe it will be needed when they are tested.

We have the generated map of `ids.ShortID:uint64` in validators_cache_test as well, and also since this map is used for inserting the cache in the type, it is also wrapped further in transitioner and manager types.

I am not sure what do you mean here - to remove all of my testing funcs from tests/mocks and simply use hardcoded maps?

plugin/evm/validators_normalizer_test.go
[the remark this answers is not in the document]
AnNiran: Okay, so your suggestion is to remove all functionality that gives random values for tests and only use hardcoded maps, is this correct ?

tests/mocks/validators.go
Ullaakut: My suggestion is:

* the only func we need to have in the `testing/mocks` package IMO is the one to generate short IDs
* validator maps should be hardcoded in existing tests
	* except in the validator normalizer tests where we should have a function to generate a random validator map (with random IDs and random weights) but that function should remain in `validators_normalizer_test.go` and not be in a package meant to be used externally

Does that make sense?

plugin/evm/validators_normalizer_test.go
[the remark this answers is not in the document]
Ullaakut: See https://github.com/flare-foundation/coreth/pull/30#discussion_r836074758

IMO it still makes sense to use random values, but only in the test that checks that given any map of validators to weight we get the right map of validators to weight ratios

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		expected, _ := mock.ByEpoch(epoch)
+
+		valCache := NewValidatorsCache(mock, WithCacheSize(32))
+		for k, v := range testValidators {
+			valCache.cache.Add(k, v)
+		}
+		result, err := valCache.ByEpoch(epoch)
+		require.NoError(t, err)
+		assert.InDeltaMapValues(t, result, expected, 0, 0)
```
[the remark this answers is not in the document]
AnNiran: This is fixed 
https://github.com/flare-foundation/coreth/pull/30/commits/7a613a2999ab621977f131c5110ca415950f16f6#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R56
Thanks!

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		expected, _ := mock.ByEpoch(epoch)
+
+		valCache := NewValidatorsCache(mock, WithCacheSize(32))
+		for k, v := range testValidators {
+			valCache.cache.Add(k, v)
+		}
+		result, err := valCache.ByEpoch(epoch)
+		require.NoError(t, err)
+		assert.Len(t, result, 0)
```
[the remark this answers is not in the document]
AnNiran: This is fixed https://github.com/flare-foundation/coreth/pull/30/commits/7a613a2999ab621977f131c5110ca415950f16f6#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R83
Thanks!

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,135 @@
...
+	})
+
+	t.Run("handles empty validators map", func(t *testing.T) {
+		var (
+			epoch          uint64 = 0
+			testValidators        = mocks.ValidatorsTestData{
+				epoch: map[ids.ShortID]uint64{},
+			}
+			calls int
+		)
```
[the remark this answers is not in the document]
AnNiran: Corrected here https://github.com/flare-foundation/coreth/pull/30/commits/7a613a2999ab621977f131c5110ca415950f16f6#diff-3bb410a6d1f259ada3678375ee08a83bd2a8bf2e93945e5c92b92a6f9d99e796R94-R103
and here also
https://github.com/flare-foundation/coreth/pull/30/commits/7a613a2999ab621977f131c5110ca415950f16f6#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7L90-R97
Thanks!

tests/mocks/validators.go
AnNiran: This function has been renamed: https://github.com/flare-foundation/coreth/pull/30/commits/d5cd7cf69b17057e500dbd2887a4f2c5122d7184
Thanks!

tests/mocks/validators.go
AnNiran: In terms of explanation yes.
`type ValidatorsTestData map[interface{}]map[ids.ShortID]uint64` is a test type that corresponds to the map type used by validators manager, transitioner, ftso and normalizer.
The function generating the validators tests data by epoch can also potentially be used in multiple places for testing.
So, if we leave only the function `RandID()` inside tests/mocks, I think the other testing data should still be placed in a separate helper/testing or whatever file or sub-package inside pluging/evm.

Also, it seems to me that this reorganization is not related to the PR anymore, what do you think of bringing it to another one?
Also it might be better to have a discussion outside the github comments and decide how to move forward first, then place the decided description in an issue and move to another PR solving it, what is your opinion about this?

plugin/evm/validators_normalizer.go
```diff
@@ -44,20 +44,24 @@ func (v *ValidatorsNormalizer) ByEpoch(epoch uint64) (map[ids.ShortID]uint64, er
...
 	}
 
+	return v.calcWeightRatio(validators), nil
+}
+
+func (v *ValidatorsNormalizer) calcWeightRatio(validators map[ids.ShortID]uint64) map[ids.ShortID]uint64 {
 	var totalWeight uint64
-	for _, weight := range validators {
-		totalWeight += weight
+	for _, wght := range validators {
```
emailtovamos, line 52: Why `weight` to `wght`?

tests/mocks/validators.go
Ullaakut: This PR is the one that adds this code to generate the validator maps, so IMO it should be solved in this PR before it gets merged 😅 If you really want to, we can merge this one as-is and open an issue about it to deal with it later, but I don't really see what is unclear/unsure at the moment 🤔 Maybe we should just get on a call to agree on a solution?

tests/mocks/validators.go
Ullaakut: `GenerateValidatorsTestData` is wayyyy too long IMO 😄 It's a detail but I'd rather have smth like `GenericValidators`. Since it is only used in testing packages, it's implicitly obvious that it will return test data, I don't think that needs to be in the function name :)

Now, regarding what we said before, since IMO this should not be in `testing/mocks` but in `plugin/evm` instead, it could even become a private function since it won't be used externally

tests/mocks/validators.go
AnNiran: I was speaking about the reorganization of the files only.
I thought it appeared as a question along with the work here, that is why I suggested it.
But of course when you have time, I would be happy to discuss it, and clear it out here also.

tests/mocks/validators.go
AnNiran: Yes, all right

plugin/evm/validators_normalizer.go
AnNiran: Because I wanted it to be shorter

plugin/evm/validators_normalizer_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+		for i := 0; i < len(validatorsData); i++ {
+			expected := mock.CalcWeightRatioFunc(testValidatorsCopy[uint64(i)])
+			result, err := valNormalizer.ByEpoch(uint64(i))
+			require.NoError(t, err)
+
+			assert.ElementsMatch(t, expected, result)
+		}
+		assert.Equal(t, 2, calls)
+	})
```
Ullaakut, line 62: I think there's still one thing that I haven't managed to explain properly :p Maybe we can have another call if it's unclear, but here is how I would have implemented this test:

```suggestion
	t.Run("nominal case", func(t *testing.T) {
		testValidators := validatorsData
		testWeightRatios := map[ids.ShortID]uint64{
			ids.ShortID{1}: 13,
			ids.ShortID{2}: 37,
		}
		
		var calls int
		mock := TestValidatorsNormalizer{
			ByEpochFunc: func(e uint64) (map[ids.ShortID]uint64, error) {
				return testValidators[e], nil
			},
			CalcWeightRatioFunc: func(vals map[ids.ShortID]uint64) map[ids.ShortID]uint64 {
				return testWeightRatios
			},
		}

		valCache := NewValidatorsCache(mock, WithCacheSize(128))
		for k, v := range testValidators {
			valCache.cache.Add(k, v)
		}
		valNormalizer := NewValidatorsNormalizer(&logging.Log{}, valCache)

		for i := 0; i < len(validatorsData); i++ {
			got, err := valNormalizer.ByEpoch(uint64(i))
			require.NoError(t, err)

			assert.ElementsMatch(t, testWeightRatios, got)
		}
		assert.Equal(t, 2, calls)
	})
```

* We no longer rely on the mock to give us the expected value, it's simply hardcoded
* We no longer modify the map we receive in the mock
* We no longer need the `duplicate` helper to duplicate the validator map
* Also some small style changes which are not very important
	* renamed result to got for consistency with other tests
	* removed `var` block for clarity

WDYT? If this seems good to you, feel free to apply those suggestions to other tests as well ^^

Ullaakut (commented):
Last comment and then this PR LGTM :D Good work!

plugin/evm/validators_normalizer_test.go
AnNiran: Corrected here https://github.com/flare-foundation/coreth/pull/30/commits/92d487996acec507b6bb402cb312a45b734f4ec9
Thanks very much for the explanation, makes sense!

Ullaakut: As I said though, if this is good for you, those suggestions should also be applied to the other subtests of this function, and then the `duplicate` helper can even be removed entirely 👍

AnNiran: Yes, I know.
Added them here https://github.com/flare-foundation/coreth/pull/30/commits/900a081da60254efa28e6b184e37f1dd40a1d54f
Thanks!

plugin/evm/validators_cache_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+		testValidators := genericValidators(numValidators, numEpochs)
+
+		var calls int
+		mock := TestValidatorCache{
+			ByEpochFunc: func(e uint64) (map[ids.ShortID]uint64, error) {
+				return testValidators[e], nil
+			},
+		}
+		expected, _ := mock.ByEpoch(epoch)
```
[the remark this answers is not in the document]
AnNiran: Yes :)
Fixed here https://github.com/flare-foundation/coreth/pull/30/commits/5f4de0f135118da86d67fdf74d4984dc01f72add#diff-53d507a9b6fbf6d169f0b9686ad727450a293914f998f4a17c524fa84623c4d7R31-R54
Thanks!

PR #31, opened 2022-03-26
Add validator transitioner tests

Ullaakut:
## Goal of this PR

This PR brings to master one commit that got lost on the now unused [`validators-ftso-transition`](https://github.com/flare-foundation/coreth/tree/validators-ftso-transition) branch.

PRs were based on it, and when it got merged I didn't realize immediately and I merged PR #24 into it, so it never got to master. This PR fixes this mistake.

It includes a single commit, by @emailtovamos

### DIFF
--- a/plugin/evm/validators_transitioner_test.go
+++ b/plugin/evm/validators_transitioner_test.go
@@ -1,0 +1,163 @@
+package evm
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/flare-foundation/flare/ids"
+)
+
+type testValidatorRetriever struct {
+	ByEpochFunc func(epoch uint64) (map[ids.ShortID]uint64, error)
+}
+
+func (t *testValidatorRetriever) ByEpoch(epoch uint64) (map[ids.ShortID]uint64, error) {
+	return t.ByEpochFunc(epoch)
+}
+
+func TestValidatorsTransitioner_ByEpoch(t *testing.T) {
+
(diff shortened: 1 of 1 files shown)


plugin/evm/validators_transitioner_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+		validatorsTransitioner := NewValidatorsTransitioner(nil, mock)
+
+		got, err := validatorsTransitioner.ByEpoch(epoch)
+		require.NoError(t, err)
+		assert.Equal(t, providers, got)
+	})
+}
+
+func fakeProviders(epoch uint64) map[ids.ShortID]uint64 {
```
AnNiran, line 144: Inside tests/mocks/validators.go, there is a specific function `func RetrieverMock(numValidators, epochs int) ValidatorsTestData` that creates random validators set with epochs.
You can simply use it by giving `1` as epoch number and then use the 0th position of the map that is received, 
as data material used across these tests.
Or, use the hardcoded validators values inside `ValidatorsData`.
The functionality for adding an additional byte, in order to be 100% sure the data is not the same, can be run on any of the above results, and IMO it should also be inside tests/mocks/validators as a method implemented by `ValidatorsTestData`, as it already implements `Duplicate`.

Ullaakut: This is true on your branch, but it has not been merged yet. We can do a round of cleanup when it gets merged, to uniformize the tests and their use of mocked values 👍

AnNiran: > This is true on your branch, but it has not been merged yet. We can do a round of cleanup when it gets merged, to uniformize the tests and their use of mocked values +1
Yes

AnNiran (approved):

PR #34, opened 2022-03-26
Add core contract upgrades

awfm9:
This PR upgrades the byte code of the flare daemon and the price submitter to the new versions.


Ullaakut (approved):
Code changes look good to me but I can't verify whether the new byte codes are correct.

consensus/misc/flare_daemon.go
```diff
@@ -0,0 +1,24 @@
...
+package misc
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+
+	"github.com/flare-foundation/coreth/core/state"
+	"github.com/flare-foundation/coreth/params"
+)
+
+var flareDaemonUpgrade = CoreContractUpdate{
```
emailtovamos, line 12: Regarding nomenclature, why one has `Upgrade` and the other has `Update` ? Sorry if I missed any previous conversation regarding this if any.

consensus/misc/flare_daemon.go
```diff
@@ -0,0 +1,24 @@
...
...38103828501526126798187612518565b915050828103604084015261268e8185612518565b9695505050505050565b600060a082526126ab60a0830188612518565b6020838203818501526126be8289612518565b848103604086015287518082529092508183019082810284018301838a01865b8381101561270c57601f198784030185526126fa838351612547565b948601949250908501906001016126de565b50508681036060880152612720818a6124d5565b955050505050508260808301529695505050505050565b6001600160a01b039290921682521515602082015260400190565b6001600160a01b0392831681529116602082015260400190565b60006020825261200b6020830184612547565b6001600160c01b039290921682526001600160401b0316602082015260400190565b90815260200190565b60e01c90565b600060443d10156127c0576104cd565b600481823e6308c379a06127d482516127aa565b146127de576104cd565b6040513d600319016004823e80513d6001600160401b03816024840111818411171561280d57505050506104cd565b8284019250825191508082111561282757505050506104cd565b503d8301602082840101111561283f575050506104cd565b601f01601f1916810160200160405291505090565b6001600160a01b0381168114610a9e57600080fdfe536166654d6174683a206d756c7469706c69636174696f6e206f766572666c6f77a26469706673582212204565821f65608e6e4fdb7076f78e5f829e603a82de85005cd73fca61d73e634f64736f6c63430007060033"),
```
emailtovamos, line 14: Having a single source of truth for the contract addresses (Old and new) 
would be great so that we have to change one file only in future.

consensus/misc/price_submitter.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+
+	"github.com/flare-foundation/coreth/core/state"
+	"github.com/flare-foundation/coreth/params"
+)
+
+var priceSubmitterUpgrade = CoreContractUpdate{
+	Address: params.ValidationAddress,
```
bostjan-kovac, line 13: This should be params.SubmitterAddress

consensus/misc/flare_daemon.go
awfm9: Those are not addresses, those are contract byte code.

PR #35, opened 2022-03-27
Take into account votepower cap

awfm9:
This PR makes sure that we cap the votepower for FTSO providers at the configured cap. Otherwise, capped FTSO validators would receive less validation weight than uncapped ones, even if their performance could be better.

### DIFF
--- a/plugin/evm/ftso_snapshot.go
+++ b/plugin/evm/ftso_snapshot.go
@@ -22,6 +22,35 @@ type FTSOSnapshot struct {
 	contracts FTSOContracts
 }
 
+func (f *FTSOSnapshot) Cap() (float64, error) {
+
+	supply := &big.Int{}
+	err := BindEVM(f.system.blockchain).
+		AtBlock(f.start).
+		OnContract(f.contracts.WNAT).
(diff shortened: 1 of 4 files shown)


plugin/evm/ftso_snapshot.go
```diff
@@ -22,6 +22,35 @@ type FTSOSnapshot struct {
...
+	if err != nil {
+		return 0, fmt.Errorf("could not get total supply: %w", err)
+	}
+
+	fraction := &big.Int{}
+	err = BindEVM(f.system.blockchain).
+		AtBlock(f.start).
+		OnContract(f.contracts.Manager).
+		Execute(Settings).
+		Decode(&fraction, nil, nil, nil, nil, nil, nil, nil, nil, nil)
```
bogdanprodanj, line 42: "settings" function returns 9 parameters 
https://songbird-explorer.flare.network/address/0xbfA12e4E1411B62EdA8B035d71735667422A6A9e/contracts

```suggestion
		Decode(&fraction, nil, nil, nil, nil, nil, nil, nil, nil)
```

PR #43, opened 2022-03-30
Moved state connector logic into separate component

bogdanprodanj:


core/state_connector_test.go
```diff
@@ -0,0 +1,268 @@
...
+func TestStateTransition_FinalisePreviousRound(t *testing.T) {
+	currentRoundNumber := []byte("222")
+	testCoinbaseAddress := common.BigToAddress(big.NewInt(1000))
+	blockTime := big.NewInt(time.Date(2022, time.February, 28, 17, 1, 1, 0, time.UTC).Unix())
+
+	t.Run("coston chain", func(t *testing.T) {
+		t.Parallel()
+
+		t.Run("nominal case", func(t *testing.T) {
+			t.Parallel()
```
Ullaakut, line 82: Not sure if this is legal 🤔 does it work as expected?

I usually only make the lowest level parallel, but if this works it's fine for me as is

core/state_connector_test.go
```diff
@@ -0,0 +1,268 @@
...
+			mockMsg, mockSCC := buildStateConnectorMock("coston chain", testCoinbaseAddress)
+
+			mockSCC.CallFunc = func(context vm.BlockContext, caller vm.ContractRef, addr common.Address, input []byte, gas uint64, value *big.Int) (ret []byte, leftOverGas uint64, err error) {
+				if caller.Address() == costonDefaultAttestors[0] {
+					rootHash := []byte("some_hash")
+					return rootHash, 0, nil
+				}
+
+				if caller.Address() != stateConnectorCoinbaseSignalAddr(params.CostonChainID, blockTime) {
+					return nil, 0, errors.New("caller address should be state connector coinbase signal addr")
```
Ullaakut, line 93: No need to return a real error since this is a mock. Or do we need this specific error to be returned for the test to succeed?

core/state_connector_test.go
```diff
@@ -0,0 +1,268 @@
...
+				}
+
+				if context.Coinbase == testCoinbaseAddress {
+					return nil, 0, errors.New("original coinbase address should have been changed to state connector coinbase signal addr")
+				}
+				return nil, 0, nil
+			}
```
Ullaakut, line 100: So this mock is called at least 4 times to hit those 4 paths in this specific subtest? 

If not, unnecessary logic should be removed.

If yes, I think it would be clearer to add comments to explain which condition is for which call, wdyt?

core/state_connector_test.go
```diff
@@ -0,0 +1,268 @@
...
```
Ullaakut, line 125: Don't we only need a single return value here, without any logic?

core/state_connector_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+			c := newConnector(mockSCC, mockMsg)
+			err := c.finalizePreviousRound(params.CostonChainID, blockTime, currentRoundNumber)
+			require.NoError(t, err)
```
Ullaakut, line 129: How is the error handled here? Seems ignored to me if we require no error return

core/state_connector.go
```diff
@@ -22,195 +21,237 @@ const (
...
+	}
 	switch {
 	case chainID.Cmp(params.CostonChainID) == 0:
 		return costonDefaultAttestors
 	case chainID.Cmp(params.SongbirdChainID) == 0:
 		return songbirdDefaultAttestors
 	case chainID.Cmp(params.FlareChainID) == 0:
 		return flareDefaultAttestors
+	default:
+		return nil
```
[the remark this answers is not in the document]
bogdanprodanj: No, since the empty attestors' list won't affect the finalization round.

core/state_connector.go
```diff
@@ -22,195 +21,237 @@ const (
...
-func (st *StateTransition) FinalisePreviousRound(chainID *big.Int, timestamp *big.Int, currentRoundNumber []byte) error {
-	getAttestationSelector := GetAttestationSelector(chainID, timestamp)
-	instructions := append(getAttestationSelector[:], currentRoundNumber[:]...)
-	defaultAttestors := GetDefaultAttestors(chainID)
-	defaultAttestationVotes, err := st.CountAttestations(defaultAttestors, instructions)
-	if err != nil {
-		return err
+// attestationResult returns resulting hash from the attestor.
+func (c *stateConnector) attestationResult(attestor common.Address, instructions []byte) (string, error) {
+	rootHash, _, err := c.caller.Call(vm.AccountRef(attestor), c.to(), instructions, 20000, big.NewInt(0))
```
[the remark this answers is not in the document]
bogdanprodanj: @awfm9, how should this value be called in the `params` package?

core/state_connector_test.go
bogdanprodanj: yes, It works :)

core/state_connector_test.go
bogdanprodanj: Yes, this error is specified to ensure that `defer` works as expected

core/state_connector_test.go
bogdanprodanj: `finalizePreviousRound` only checks for error and omits other

core/state_connector_test.go
bogdanprodanj: Round finalization should not be affected when one of the attestors returns an error, thus `require.NoError(t, err)`

PR #44, opened 2022-03-30
evm: fix validators unit tests.

AnNiran:


plugin/evm/validators_cache_test.go
```diff
@@ -100,13 +99,11 @@ func TestValidatorsCache_ByEpoch(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Empty(t, got)
 		assert.ElementsMatch(t, testValidators[epoch], got)
```
emailtovamos, line 101: How is `assert.ElementsMatch` different from `assert.Equal` in the case?

plugin/evm/validators_normalizer_test.go
```diff
@@ -23,18 +23,14 @@ func (t TestValidatorsNormalizer) ByEpoch(e uint64) (map[ids.ShortID]uint64, err
 func TestValidatorsNormalizer_ByEpoch(t *testing.T) {
 	t.Run("nominal case", func(t *testing.T) {
 		testValidators := validatorsData
-		testWeightRatios := map[ids.ShortID]uint64{
-			{1}: 13,
-			{2}: 37,
-		}
+		testWeightRatios := validatorsData
```
emailtovamos, line 26: I didn't fully understand what was wrong in this case before. Can you explain?

plugin/evm/validators_cache_test.go
```diff
@@ -37,7 +37,6 @@ func TestValidatorsCache_ByEpoch(t *testing.T) {
 
 		testValidators := genericValidators(numValidators, numEpochs)
 
-		var calls int
```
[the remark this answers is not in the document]
AnNiran: The calls was removed because it was following the number of times the mock is called.
After the last changes before the merge, the tests were set to not call the mock anymore, but only compare hardcoded values with the result from calling the tested function.
That is why I removed it - because the mock is not called anymore.

plugin/evm/validators_cache_test.go
AnNiran: assert.ElementsMatch works successfully with lists - arrays or slices, not maps.

plugin/evm/validators_normalizer_test.go
AnNiran: This test uses a hardcoded initial value:
```
validatorsData = validatorsTestData{
		uint64(0): map[ids.ShortID]uint64{
			ids.ShortID([20]byte{202, 22, 105, 80, 240, 156, 96, 235, 127, 79, 82, 60, 159, 12, 172, 216, 171, 131, 235, 63}):   6185718675989118760,
			ids.ShortID([20]byte{182, 115, 39, 234, 78, 181, 63, 96, 108, 42, 134, 160, 148, 157, 103, 198, 77, 146, 118, 200}): 6465383072520688749,
			ids.ShortID([20]byte{20, 206, 223, 250, 40, 74, 158, 159, 174, 134, 44, 20, 207, 134, 220, 231, 63, 247, 84, 240}):  515007957320493220,
		},
		uint64(1): map[ids.ShortID]uint64{
			ids.ShortID([20]byte{231, 37, 78, 169, 39, 173, 150, 37, 20, 7, 91, 212, 235, 6, 235, 74, 120, 22, 233, 150}):     9599604527249266409,
			ids.ShortID([20]byte{236, 230, 116, 165, 56, 37, 35, 211, 216, 60, 90, 243, 59, 36, 173, 140, 109, 252, 169, 32}): 3417212044008417831,
		},
	}```

Each of the keys have zero weight after ratio calculation inside `calcWeightRatio`.
in order to be compared for testing, the map that it is compared against need to have same keys with zero values, not the previously existing map
```
testWeightRatios := map[ids.ShortID]uint64{
			{1}: 13,
			{2}: 37,
		}
```

plugin/evm/validators_cache_test.go
emailtovamos: You mean assert.Equal in the above sentence?

AnNiran: No, 
I meant `assert.ElementsMatch`.
About the `assert.Equal` fails randomly when comparing maps,
and I prefer to test the maps either using reflect.DeepEqual() or inspect their keys and length myself.
Here I used the second one.

PR #48, opened 2022-03-31
Fixed data provider gas refund

### LINKED ISSUE #45: Fix data provider gas refund
The state connector has a mechanism to refund data providers for transactions to the price provider, if the transactions succeed.

This used the nominal fee for a normal transaction of tokens as a baseline, which was 21k gas with a gas price of 225M. However, with the hard fork we did, the minimum gas price was reduced to 25M instead. This means that data provider submissions are now 9 times as expensive as they should be.

We need to implement a mechanism to transition to the new gas price as part of the next hard fork.

bogdanprodanj:
Fix #45

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -386,6 +386,11 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	if vmerr == nil && msg.To() != nil && *msg.To() == common.HexToAddress(GetPrioritisedFTSOContract(timestamp)) {
 		nominalGasUsed := uint64(21000)
 		nominalGasPrice := uint64(225_000_000_000)
+
+		if st.evm.ChainConfig().IsFlareHardFork1(st.evm.Context.Time) {
+			nominalGasPrice = uint64(25_000_000_000)
+		}
+
 		nominalFee := new(big.Int).Mul(new(big.Int).SetUint64(nominalGasUsed), new(big.Int).SetUint64(nominalGasPrice))
 		actualGasUsed := st.gasUsed()
 		actualGasPrice := st.gasPrice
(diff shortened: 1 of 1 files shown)


core/state_transition.go
```diff
@@ -386,6 +386,11 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	if vmerr == nil && msg.To() != nil && *msg.To() == common.HexToAddress(GetPrioritisedFTSOContract(timestamp)) {
 		nominalGasUsed := uint64(21000)
 		nominalGasPrice := uint64(225_000_000_000)
+
+		if st.evm.ChainConfig().IsFlareHardFork1(st.evm.Context.Time) {
```
awfm9, line 390: Did you check that this corresponds to the timestamp in the block that is being processed?

core/state_transition.go
```diff
@@ -386,6 +386,11 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	if vmerr == nil && msg.To() != nil && *msg.To() == common.HexToAddress(GetPrioritisedFTSOContract(timestamp)) {
 		nominalGasUsed := uint64(21000)
 		nominalGasPrice := uint64(225_000_000_000)
```
awfm9, line 388: We can use `params.ApricotPhase1MinGasPrice` here, to be more explicit.

core/state_transition.go
```diff
@@ -386,6 +386,11 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
+
+		if st.evm.ChainConfig().IsFlareHardFork1(st.evm.Context.Time) {
+			nominalGasPrice = uint64(25_000_000_000)
```
awfm9, line 391: Then we can use `params.ApricotPhase4MinBaseFee` here, to also be more explicit.

awfm9 (commented):
Seems like it was easier to fix than I thought :)

We can make it a bit more explanatory by using defined constants. Also, just double-checking if you are sure about the timestamp.

core/state_transition.go
bogdanprodanj: Yes, data from the block is set into evm context.
```
blockContext := NewEVMBlockContext(header, p.bc, nil)
vmenv := vm.NewEVM(blockContext, vm.TxContext{}, statedb, p.config, cfg)
```

Issue #46, opened 2022-03-31
Increase maximum mint request


awfm9:
Currently, we can only mint a maximum of 50M, as dictated by the Flare Daemon. We need to increase this limit to 90M instead of 50M. It's given in the `GetMaximumMintRequest` function in `keeper.go`.

Like the other changes, this needs to activate as part of the upcoming hard fork.

awfm9:
Done in #50.

PR #50, opened 2022-04-01
Increase max mint request after hardfork. 

emailtovamos:
For https://github.com/flare-foundation/coreth/issues/46

The timestamp seems fine but remains to be confirmed.

TODO: fix broken tests

### DIFF
--- a/core/keeper.go
+++ b/core/keeper.go
@@ -115,13 +115,13 @@ func triggerKeeper(evmCaller EVMCaller) (*big.Int, error) {
 	}
 }
 
-func mint(evm EVMCaller, mintRequest *big.Int, stEvm *vm.EVM) error {
+func mint(evmCaller EVMCaller, mintRequest *big.Int, evm *vm.EVM) error {
 	// If the mint request is greater than zero and less than max
-	max := GetMaximumMintRequest(evm.GetBlockNumber(), stEvm)
+	max := GetMaximumMintRequest(evmCaller.GetBlockNumber(), evm)
 	if mintRequest.Cmp(big.NewInt(0)) > 0 &&
 		mintRequest.Cmp(max) <= 0 {
 		// Mint the amount asked for on to the keeper contract
-		evm.AddBalance(common.HexToAddress(GetSystemTriggerContractAddr(evm.GetBlockNumber())), mintRequest)
+		evmCaller.AddBalance(common.HexToAddress(GetSystemTriggerContractAddr(evmCaller.GetBlockNumber())), mintRequest)
 	} else if mintRequest.Cmp(max) > 0 {
 		// Return error
 		return &ErrMaxMintExceeded{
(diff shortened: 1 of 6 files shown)


core/keeper.go
```diff
@@ -72,24 +72,26 @@ func GetPrioritisedFTSOContract(blockTime *big.Int) string {
 	}
 }
 
-func GetMaximumMintRequest(blockNumber *big.Int) *big.Int {
-	switch {
-	default:
-		maxRequest, _ := new(big.Int).SetString("50000000000000000000000000", 10)
-		return maxRequest
+func GetMaximumMintRequest(blockNumber *big.Int, evm *vm.EVM) *big.Int {
```
awfm9, line 75: As we are not using the block number to switch here, but rather the block time, the signature should simply switch to this:

```go
func GetMaximumMintRequest(blockTime *big.Int) *big.Int
```

There is no reason to inject the _entire_ EVM just for a single value.

core/keeper.go
```diff
@@ -72,24 +72,26 @@ func GetPrioritisedFTSOContract(blockTime *big.Int) string {
...
+
+	if evm.ChainConfig().IsFlareHardFork1(evm.Context.Time) {
+		maxRequest, _ = new(big.Int).SetString("90000000000000000000000000", 10)
 	}
+
+	return maxRequest
 }
 
-func triggerKeeper(evm EVMCaller) (*big.Int, error) {
+func triggerKeeper(evmCaller EVMCaller) (*big.Int, error) {
```
awfm9, line 85: `EVMCaller` is an interface that represents the EVM already. There is no need to rename it, because the concrete underlying type is the EVM.

core/keeper.go
```diff
@@ -113,13 +115,13 @@ func triggerKeeper(evm EVMCaller) (*big.Int, error) {
 	}
 }
 
-func mint(evm EVMCaller, mintRequest *big.Int) error {
+func mint(evmCaller EVMCaller, mintRequest *big.Int, evm *vm.EVM) error {
```
awfm9, line 118: Having both `EVMCaller` and `*vm.EVM` makes no sense here. They are both the same underlying value. Instead, you should expose the related function of the `*vm.EVM` type by adding it to the `EVMCaller` interface. Also, the `GetBlockNumber()` is no longer required.

core/keeper.go
```diff
@@ -134,13 +136,13 @@ func mint(evm EVMCaller, mintRequest *big.Int) error {
 	return nil
 }
 
-func triggerKeeperAndMint(evm EVMCaller, log log.Logger) {
+func triggerKeeperAndMint(evmCaller EVMCaller, log log.Logger, evm *vm.EVM) {
```
awfm9, line 139: Same thing here. You are injecting two parameters that are the same underlying value. Additionally, you are breaking the loose coupling by using a concrete type instead of the interface.

awfm9 (changes requested):
Hey @emailtovamos, this has a few problems, see comments. I believe you will be travelling, so I will go ahead and fix it myself, but I added the comments anyway so you can review them.

PR #51, opened 2022-04-01
Add default validators component

awfm9:
This PR makes the default validators component a bit more flexible, so we can phase out some default validators after the FTSO validators take over.

Once the FTSO validators fully take over for the default validators, the default validators will only be used in two cases:

1. some FTSO validators unset their Node ID and thus stop running their validators; or
2. some catastrophic glitch on the FTSO makes reward payouts zero for one reward epoch.

In the first case, this change will ensure that default validators are only added back into the mix if the number falls much lower. So the network will keep running with 18, or even just 6, FTSO validators without falling back to centralized default validators.

In the second case, the network will fall back only on 3 or 5 default validators in case of a catastrophic glitch, which means we have to maintain less default validators running, reducing costs. Additionally, the transition away from default validators to FTSO  will happen much faster in that case.

The PR also makes it possible to phase out multiple default validators per reward epoch, and sets the values so that we reach an active validator set of mostly FTSO validators around Flare main network launch.

### DIFF
--- a/plugin/evm/validators_default.go
+++ b/plugin/evm/validators_default.go
@@ -1,9 +1,11 @@
 package evm
 
 import (
+	"bytes"
 	"fmt"
 	"math/big"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/flare-foundation/flare/ids"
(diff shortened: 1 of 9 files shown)


plugin/evm/validators_default.go
```diff
@@ -52,19 +54,37 @@ var songbirdNodeIDs = []string{
...
+func NewValidatorsDefault(chainID *big.Int) (*ValidatorsDefault, error) {
 
 	var weight uint64
 	var nodeIDs []string
+	var steps []Step
 	switch {
 	case chainID.Cmp(params.CostonChainID) == 0:
 		nodeIDs = costonNodeIDs
 		weight = costonValidatorWeight
+		steps = []Step{
```
bogdanprodanj, line 76: Could you please put names for these epochs in the comments?

plugin/evm/validators_default.go
```diff
@@ -90,5 +110,40 @@ func getDefaultValidators(chainID *big.Int) (map[ids.ShortID]uint64, error) {
...
+}
+
+func (v *ValidatorsDefault) ByEpoch(epoch uint64) (map[ids.ShortID]uint64, error) {
+
+	validatorIDs := make([]ids.ShortID, 0, len(v.validators))
+	for validatorID := range v.validators {
+		validatorIDs = append(validatorIDs, validatorID)
+	}
+	sort.Slice(validatorIDs, func(i int, j int) bool {
+		return bytes.Compare(validatorIDs[i][:], validatorIDs[j][:]) < 0
```
AnNiran, line 128: This line outputs a boolean value

AnNiran (approved):

plugin/evm/validators_default.go
awfm9: I don't know what you mean?

PR #55, opened 2022-04-05
Fix tests for validators default

emailtovamos:


plugin/evm/validators_default_test.go
```diff
@@ -47,13 +44,13 @@ func TestValidatorsDefault_ByEpoch(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			t.Parallel()
 
-			got, err := getDefaultValidators(test.chainID)
+			got, err := NewValidatorsDefault(test.chainID)
 			test.wantErr(t, err)
 
 			if err == nil {
```
bogdanprodanj, line 50: You can skip `if err == nil {` here since `require..*` will fail before this check.

plugin/evm/validators_default_test.go
```diff
@@ -66,13 +63,62 @@ func TestGetDefaultValidators_CustomValidators(t *testing.T) {
 	err := os.Setenv("CUSTOM_VALIDATORS", "NodeID-3M9KVT6ixi4gVMisbm5TnPXYXgFN5LHuv,NodeID-NnX4fajAmyvpL9RLfheNdc47FKKDuQW8i")
```
bogdanprodanj, line 63: You should probably change the name of this test case.
TestNewValidatorsDefault -> TestNewValidatorsDefault_StandardNetwork
TestGetDefaultValidators_CustomValidators -> TestNewValidatorsDefault_CustomNetwork

wdyt?

plugin/evm/validators_normalizer.go
```diff
@@ -48,20 +49,20 @@ func (v *ValidatorsNormalizer) ByEpoch(epoch uint64) (map[ids.ShortID]uint64, er
...
+
 	var totalWeight uint64
-	for _, wght := range validators {
-		totalWeight += wght
+	for _, weight := range validators {
+		totalWeight += weight
 	}
 
-	v.log.Debug("normalizing weight from %d to %d", totalWeight, math.MaxInt32)
+	v.log.Debug("normalizing weight total from %d to %d", totalWeight, math.MaxInt32)
```
bogdanprodanj, line 58:
```suggestion
	v.log.Debug("normalizing total weight from %d to %d", totalWeight, math.MaxInt32)
```

PR #56, opened 2022-04-05
Fix validators transitioner tests

emailtovamos:


plugin/evm/validators_transitioner_test.go
```diff
@@ -20,12 +21,14 @@ func (t *testValidatorRetriever) ByEpoch(epoch uint64) (map[ids.ShortID]uint64,
 
 func TestValidatorsTransitioner_ByEpoch(t *testing.T) {
 
-	t.Run("nominal case 1: includes check for recursion and cache usage", func(t *testing.T) {
+	t.Run("nominal case 1: includes check for recursion and number of calls in the absence of cache and compare weights", func(t *testing.T) {
 		t.Parallel()
 
 		calls := 0
 		testEpoch := uint64(10)
+		wantWeight := uint64(45)
```
bogdanprodanj, line 29: You should either use `want*` values in all asserts, or just plain numbers

bogdanprodanj (approved):

plugin/evm/validators_transitioner_test.go
```diff
@@ -3,164 +3,214 @@
...
+		providers[ids.ShortID{byte(i)}] = uint64(i)
+	}
+
+	return providers
+}
+
+// fakeValidators gives a different set of validators compared to fakeProviders() to have no overlap for now
+func fakeValidators(epoch uint64) map[ids.ShortID]uint64 {
+
+	providers := make(map[ids.ShortID]uint64)
```
awfm9, line 210: This should be named `validators` if you do it like this.

plugin/evm/validators_transitioner_test.go
```diff
@@ -3,164 +3,214 @@
...
+			}
+		})
+	}
+}
+
+func fakeProviders(epoch uint64) map[ids.ShortID]uint64 {
+
+	providers := make(map[ids.ShortID]uint64)
+	for i := 0; i < int(epoch); i++ {
+		providers[ids.ShortID{byte(i)}] = uint64(i)
```
awfm9, line 201: This means that the weight will be zero for the first validator, which means it will be ignored. It's not a valid weight.

plugin/evm/validators_transitioner_test.go
```diff
@@ -3,164 +3,214 @@
...
+
+	return providers
+}
+
+// fakeValidators gives a different set of validators compared to fakeProviders() to have no overlap for now
+func fakeValidators(epoch uint64) map[ids.ShortID]uint64 {
+
+	providers := make(map[ids.ShortID]uint64)
+	for i := 0; i < int(epoch); i++ {
+		providers[ids.ShortID{byte(i + int(epoch))}] = uint64(i)
```
awfm9, line 212: This means on epoch zero, validators and providers will return the same. It also doesn't work for other epochs with higher validator numbers. For example, on epoch 4, validators will be 0, 1, 2, 3, 4, 5, 6, ..., while providers will be 4, 5, 6, 7, ...

plugin/evm/validators_transitioner_test.go
```diff
@@ -3,164 +3,214 @@
...
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			v := &ValidatorsDefault{
+				validators: validators,
+				steps:      test.steps,
+			}
```
awfm9, line 182: This doesn't make sense. You should not test the default validators here. This is the test for the transitioner.

plugin/evm/validators_transitioner_test.go
```diff
@@ -3,164 +3,214 @@
...
+
+			transitioner := NewValidatorsTransitioner(logging.NoLog{}, v, mockProvider, WithStepSize(1))
```
awfm9, line 184: You should inject the validators and providers mocks here, not an actual default validators.

awfm9 (changes requested):

awfm9:
This is not actually working at the moment, but I merged the refactored tests anyway because I will change the transitioner and want to avoid rewriting them again.

PR #58, opened 2022-04-06
Persist validators & bootstrap caches

awfm9:
Relying on retrieving different versions of the state trie, all the way back to the start of the FTSO, in order to properly transition validators, has two problems:

1. it could be a potential denial-of-service vector, if people manage to inject requests for the active validators for old epochs, where validator sets have already been ejected from the cache; and
2. more importantly, it could be that we are not able to go back to the full state at every single epoch in the past, because of various pruning mechanisms.

This PR adds code that will store the active validator set for every epoch on disk, so that we only need to look at the active validators from the previous epoch, as stored on disk, in order to determine the validator set for the next epoch.

After the PR, we still support recursing all the way to the zero epoch, if need be. This is helpful in case some of the DB was manually deleted or corrupted, and allows us to rebuild the validators store. But in general, the transitioner should only be called now when we built the set for a new epoch.


plugin/evm/vm.go
```diff
@@ -445,23 +451,29 @@ func (vm *VM) Initialize(
...
+
+	// Getting the active validators for the current epoch will bootstrap all of the
+	// storage that is part of the transition logic, and will avoid long delays once
+	// we start processing blocks.
+	epoch, err := ftso.Current(lastAccepted.Hash())
+	if err != nil && !errors.Is(err, errFTSONotDeployed) && !errors.Is(err, errFTSONotActive) {
+		return fmt.Errorf("could not get current epoch: %w", err)
+	}
+	if err == nil {
+		_, err = normalizeTransitionValidators.ByEpoch(epoch)
```
emailtovamos, line 467: I didn't understand what exactly is happening here.

plugin/evm/validators_store.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return fmt.Errorf("could not put validator data: %w", err)
+	}
+
+	v.log.Debug("persisted validators for epoch %d", epoch)
+
+	return nil
+}
+
+func (v *ValidatorsStore) ByEpoch(epoch uint64) (map[ids.ShortID]uint64, error) {
```
emailtovamos, line 84: Where does this function get used? I couldn't find.

awfm9: In the validators cache that wraps around the store.

plugin/evm/vm.go
awfm9: It's described in the comment. It will execute the transitioner code, which will populate the validators store and all of the intermediate caches.

PR #60, opened 2022-04-12
Moved some of the flare config to params package

bogdanprodanj:


params/flare.go
```diff
@@ -0,0 +1,268 @@
...
+	"fmt"
+	"math/big"
+	"os"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/flare-foundation/flare/ids"
+	"github.com/flare-foundation/flare/utils/constants"
```
Ullaakut, line 14: Should be split

params/flare.go
```diff
@@ -0,0 +1,268 @@
...
+
+	v := ValidatorsDefault{
+		validators: validators,
+		steps:      steps,
+	}
+
+	return &v, nil
+}
+
+type ValidatorsDefault struct {
```
emailtovamos, line 230: I am not sure why the business logic related to `ValidatorsDefault` had to move to `params` folder. I would have expected mostly the hardcoded stuff only to move in?

PR #61, opened 2022-04-25
Added tests to cover contract updates

bogdanprodanj:


core/state_processor_test.go
```diff
@@ -354,3 +359,51 @@ func GenerateBadBlock(parent *types.Block, engine consensus.Engine, txs types.Tr
 	// Assemble and return the final block for sealing
 	return types.NewBlock(header, txs, nil, receipts, trie.NewStackTrie(nil), nil, true)
 }
+
+func TestStateProcessor_Process(t *testing.T) {
+	database := rawdb.NewMemoryDatabase()
```
Ullaakut, line 364: No need to defer closing it or smth like that?

core/state_processor_test.go
```diff
@@ -354,3 +359,51 @@ func GenerateBadBlock(parent *types.Block, engine consensus.Engine, txs types.Tr
...
+	}
+
+	cpcfg := params.TestChainConfig
+	cpcfg.ChainID = params.TestingChainID
+	genesis := Genesis{Config: cpcfg, GasLimit: gasLimit, Alloc: alloc, Coinbase: common.HexToAddress("0x0100000000000000000000000000000000000000")}
+	b := genesis.MustCommit(database)
+	cacheConfig := &CacheConfig{}
+	blockchain, _ := NewBlockChain(database, cacheConfig, genesis.Config, dummy.NewFaker(), vm.Config{}, common.Hash{})
+
+	chain, _, err := GenerateChain(genesis.Config, b, dummy.NewFaker(), database, 3, 10, func(i int, gen *BlockGen) {})
```
Ullaakut, line 386: Why 3 and 10?

core/state_processor_test.go
```diff
@@ -354,3 +359,51 @@ func GenerateBadBlock(parent *types.Block, engine consensus.Engine, txs types.Tr
...
+	balance := big.NewInt(0).SetUint64(math.MaxUint64)
+	alloc := make(GenesisAlloc)
+	alloc[auth.From] = GenesisAccount{
+		Balance: balance,
+	}
+
+	cpcfg := params.TestChainConfig
+	cpcfg.ChainID = params.TestingChainID
+	genesis := Genesis{Config: cpcfg, GasLimit: gasLimit, Alloc: alloc, Coinbase: common.HexToAddress("0x0100000000000000000000000000000000000000")}
+	b := genesis.MustCommit(database)
```
Ullaakut, line 382: this line should be moved below IMO, since it's needed for 386 and not related to the lines directly above and below it

miner/worker_test.go
```diff
@@ -0,0 +1,124 @@
+package miner
```
Ullaakut, line 1: These kind of look like integration tests, shouldn't we run them under a build tag and shouldn't they be in a separate package? (miner_test)

emailtovamos (commented):
Since this test is mainly checking the hardfork section of the `Process()` function, maybe this test function can be named differently specifying its purpose?

core/state_processor_test.go
```diff
@@ -354,3 +359,51 @@ func GenerateBadBlock(parent *types.Block, engine consensus.Engine, txs types.Tr
...
+	cacheConfig := &CacheConfig{}
+	blockchain, _ := NewBlockChain(database, cacheConfig, genesis.Config, dummy.NewFaker(), vm.Config{}, common.Hash{})
+
+	chain, _, err := GenerateChain(genesis.Config, b, dummy.NewFaker(), database, 3, 10, func(i int, gen *BlockGen) {})
+	require.NoError(t, err)
+
+	block := chain[0]
+
+	parent := blockchain.GetHeader(block.ParentHash(), block.NumberU64()-1)
```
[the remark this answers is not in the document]
Ullaakut: Since we're getting the parent, I assume we have to give its hash and height yes

core/state_processor_test.go
bogdanprodanj: Block needs to be committed to the database, because database is used in creating blockchain

core/state_processor_test.go
bogdanprodanj: I defer closing the blockchain itself

core/state_processor_test.go
bogdanprodanj: I changed `3` to `1` since I only needed one block. As for 10, it's sort of a block timestamp that affects the gas limits. In our test case, it can be any. I'll change it to `0` since it doesn't affect the test outcome.

